PR #8, opened 2020-09-04
Fix UI issues

Jevan5:
- Seekbar updates every 250ms instead of 1s
- Video no longer goes past end of window
- Play/pause controlled by spacebar instead of 'k'
- No gap in video controls when fullscreen


AdmiralCoco (approved):

media-player.js
```diff
@@ -28,27 +29,28 @@ class MediaPlayer extends InternalLocalizeMixin(LitElement) {
...
+				display: block;
 				height: 100%;
+				max-height: 100vh;
 				width: 100%;
 			}
 
 			#d2l-labs-media-player-video-controls {
 				background-image: linear-gradient(to bottom, rgba(0, 0, 0, 0.5), rgba(0, 0, 0, 0.7));
-				bottom: 7px;
+				bottom: 0px;
```
dbatiste, line 45: Nit: zero value dimensions should not have units.  The style lint config will catch this sort of thing for you.

media-player.js
```diff
@@ -287,25 +306,25 @@ class MediaPlayer extends InternalLocalizeMixin(LitElement) {
 
 					<div id="d2l-labs-media-player-flex-filler"></div>
 
-					<div id="d2l-labs-media-player-speed-container" @mouseenter=${this._startHoveringSpeedContainer} @mouseleave=${this._stopHoveringSpeedContainer}>
-						<button class="d2l-labs-media-player-control-element d2l-labs-media-player-control-display d2l-labs-media-player-button">
+					<div @mouseenter=${this._startHoveringSpeedContainer} @mouseleave=${this._stopHoveringSpeedContainer}>
+						<button class="d2l-labs-media-player-control-element d2l-labs-media-player-control-display d2l-labs-media-player-button" id="d2l-labs-media-player-speed-button">
 							${this._speed}x
```
dbatiste, line 311: Is the "x" meant to mean "times".  I think there may be a code point for this.  Or, more generally, how is this announced to a screen reader user?

package.json
```diff
@@ -45,9 +45,10 @@
     "require-dir": "^1.2.0"
   },
   "dependencies": {
-    "@brightspace-ui/core": "^1",
+    "@brightspace-ui/core": "^1.76.1",
     "@d2l/seek-bar": "Brightspace/d2l-seek-bar#semver:^1",
     "@webcomponents/webcomponentsjs": "^2",
-    "lit-element": "^2"
+    "lit-element": "^2",
+    "screenfull": "^5.0.2"
```
dbatiste, line 52: Presumably `screenfull` is a library that helps make things fullscreen?  Does it make the video go "full page", or "full screen"?  Wondering if we've looked into the implementation for this... my guess is that it's probably pretty simple and we'd prefer to implement a little helper (possibly in `core`) rather than pull in a 3rd-party dependency.

PR #9, opened 2020-09-15
Us118769/basic video player

Jevan5:


media-player.js
```diff
@@ -235,150 +245,127 @@ class MediaPlayer extends InternalLocalizeMixin(LitElement) {
...
 					</div>
 
 					<div id="d2l-labs-media-player-flex-filler"></div>
 
-					<div @mouseenter=${this._startHoveringSpeedContainer} @mouseleave=${this._stopHoveringSpeedContainer}>
-						<button class="d2l-labs-media-player-control-element d2l-labs-media-player-control-display d2l-labs-media-player-button" id="d2l-labs-media-player-speed-button">
-							${this._speed}x
+					<div class="d2l-labs-media-player-control-element" @mouseenter=${this._startUsingSpeedContainer} @mouseleave=${this._stopUsingSpeedContainer}>
+						<button class="d2l-labs-media-player-control-display d2l-labs-media-player-button" id="d2l-labs-media-player-speed-button" @focus=${this._startUsingSpeedContainer} @focusout=${this._stopUsingSpeedContainer}>
+							${this._speed}
```
dlockhart, line 312: I think this is going to need some more work to make it behave like a true menu for screen readers, but we can work more on that later and engage some specialized accessibility testing.

media-player.js
```diff
@@ -235,150 +245,127 @@ class MediaPlayer extends InternalLocalizeMixin(LitElement) {
...
-	}
+		const resizeObserver = new ResizeObserver((entries) => {
+			entries.forEach(() => {
+				const backgroundHeightPx = this.offsetHeight - 60;
+				this._speedLevelBackground.style.maxHeight = `${backgroundHeightPx}px`;
+				this._speedLevelButtons.forEach((button) => {
+					button.style.maxHeight = `${Math.floor(backgroundHeightPx / 7)}px`;
+				});
+			});
+		});
```
dlockhart, line 362: What's the observer trying to accomplish? Resize observers don't perform very well, and ideally we're not adjusting the look of things outside of the `render()` method, so hopefully we can find a different way to do this.

Jevan5: Depending on how large the video container is, it resizes the playback rate selection menu.

Jevan5: Screen Shot 2020-09-16 at 1 19 28 PM
Screen Shot 2020-09-16 at 1 20 01 PM

dlockhart: Is that totally needed? Like normally menus are allowed to overflow their parents without any issue. What happens if the video is like `50px` high? `10px`? Seems like an overly complex thing to support.

Jevan5: With a larger video player, I thought the menu looked nicer with more padding between text. However, this was not possible when the video player got smaller. I also didn't think it looked as clean when it overflowed the player. Youtube, for example, reduces the height of the menu and maintains the height of the content with scroll.

This is something that I can bring up with designers.

README.md
```diff
@@ -41,9 +41,63 @@ npm install @brightspace-ui-labs/media-player
...
 
-| Property | Type | Description |
-|--|--|--|
-| src | string | URL of the media to play. |
+| Property | Type | Default | Description |
+|--|--|--|--|
+| autoplay | boolean | false | If set, will play the media as soon as it has been loaded. |
+| loop | boolean | false | If set, once the media has finished playing it will replay from the beginning. |
+| poster | string | null | URL of the image to display in place of the media before it has loaded. |
+| src | string | null | URL of the media to play. |
```
dbatiste, line 49: Is the `src` required?  If so, normally we'd list that like this:
```
| `src` | String, required | URL of the media to play |
```

media-player.js
```diff
@@ -235,173 +278,230 @@ class MediaPlayer extends InternalLocalizeMixin(LitElement) {
...
+
 		return html`
-		<div id="d2l-labs-media-player-video-container" style=${styleMap(this._videoContainerStyle)} ?hidden="${this.src === undefined}" @mousemove=${this._showControlsTemporarily}>
-			<video ?controls="${nativeControls}" id="d2l-labs-media-player-video" preload="metadata" @play=${this._onPlay} @pause=${this._onPause} @loadedmetadata=${this._onLoadedMetadata} @loadeddata=${this._onLoadedData} @click=${this._onVideoClick} @volumechange=${this._onVolumeChange}>
+		<div id="d2l-labs-media-player-video-container" style=${styleMap(videoContainerStyle)} ?hidden="${this.src === null}" @mousemove=${this._onVideoContainerMouseMove} @click=${this._videoContainerClicked} @keydown=${this._listenForKeyboard}>
+			<video ?controls="${NATIVE_CONTROLS}" id="d2l-labs-media-player-video" ?autoplay="${this.autoplay}" ?loop="${this.loop}" poster="${this.poster ? this.poster : ''}" preload="metadata" @click=${this._onVideoClick} @ended=${this._onEnded} @loadeddata=${this._onLoadedData} @play=${this._onPlay} @pause=${this._onPause} @loadedmetadata=${this._onLoadedMetadata} @timeupdate=${this._onTimeUpdate} @volumechange=${this._onVolumeChange}>
 				<source src="${this.src}">
```
dbatiste, line 336: According to [the specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source#attr-src), the attribute is required for audio and video media sources.  

This `src` property is initialized as `null` - what happens if the consumer doesn't provide a `src`.  I think when I tested this previously it was `undefined`, which resulted in a HTTP 404.  We can guard against this by leaving it as `undefined` and:

``` javascript
import { ifDefined } from 'lit-html/directives/if-defined.js';

render() {
    ...
   <video src="${ifDefined(this.src)}">
    ...
}
```
to avoid rendering an explicit `<video src="undefined"> or <video src="null">`.  Optionally, writing a console warning for the developer if the attribute is indeed required and unspecified.

dbatiste: Thinking about this more - maybe we'd actually want this scenario logged.  @dlockhart , thoughts?

dlockhart: Doing a `console.warn` in `firstUpdated` is what we've done elsewhere when we want to enforce a required attribute. [Here's an example in numeric input](https://github.com/BrightspaceUI/core/blob/master/components/inputs/input-number.js#L81).

PR #11, opened 2020-10-01
Added audio-only support

Jevan5:
The player tries to detect the type of content in the 'src'
attribute, and renders a `video` or `audio` tag based on
the type.

The audio player is much taller than most audio players seen
on the web, to support the growing size of the settings menu
and the closed captions/subtitles which will be added in the
future.


demo/index.html
```diff
@@ -15,6 +15,8 @@
 	<body>
 		<d2l-demo-snippet>
 			<d2l-labs-media-player src="http://clips.vorwaerts-gmbh.de/big_buck_bunny.mp4"></d2l-labs-media-player>
+			<br>
+			<d2l-labs-media-player src="https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3"></d2l-labs-media-player>
 		</d2l-demo-snippet>
```
dlockhart, line 20: Typically we'd break this into two separate `<d2l-demo-snippet>`s, with a heading for each:

```suggestion
		<h2>Video Player</h2>
		<d2l-demo-snippet>
			<d2l-labs-media-player src="http://clips.vorwaerts-gmbh.de/big_buck_bunny.mp4"></d2l-labs-media-player>
		</d2l-demo-snippet>

		<h2>Audio Player</h2>
		<d2l-demo-snippet>
			<d2l-labs-media-player src="https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3"></d2l-labs-media-player>
		</d2l-demo-snippet>
```

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@brightspace-ui-labs/media-player",
   "description": "A reusable media player component.",
   "repository": "https://github.com/BrightspaceUILabs/media-player.git",
-  "version": "1.5.0",
+  "version": "1.6.0",
```
dlockhart, line 5: At some point I'd suggest setting up [semantic-release](https://semantic-release.gitbook.io/semantic-release/) to automatically bump versions and publish to NPM for you. I can help with that if you like.

media-player.js
```diff
@@ -327,18 +366,39 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
-			<video ?controls="${NATIVE_CONTROLS}" id="d2l-labs-media-player-video" ?autoplay="${this.autoplay}" ?loop="${this.loop}" poster="${this.poster ? this.poster : ''}" preload="metadata" @click=${this._onVideoClick} @ended=${this._onEnded} @loadeddata=${this._onLoadedData} @play=${this._onPlay} @pause=${this._onPause} @loadedmetadata=${this._onLoadedMetadata} @timeupdate=${this._onTimeUpdate} @volumechange=${this._onVolumeChange}>
-				<source src="${ifDefined(this.src)}">
+		const video = this._isVideo ? html`
+			<video ?controls="${NATIVE_CONTROLS}" id="d2l-labs-media-player-video" ?autoplay="${this.autoplay}" ?loop="${this.loop}" poster="${ifDefined(this.poster)}" preload="metadata" @click=${this._onVideoClick} @ended=${this._onEnded} @loadeddata=${this._onLoadedData} @play=${this._onPlay} @pause=${this._onPause} @loadedmetadata=${this._onLoadedMetadata} @timeupdate=${this._onTimeUpdate} @volumechange=${this._onVolumeChange}>
+				<source src="${this.src}">
 			</video>
+		` : null;
+
```
dlockhart, line 378: I'd suggest moving this into a private `_getContentAreaView()` method that returns the correct thing based on whether it's audio or video.

```javascript
_getContentAreaView() {
    if (this._isVideo) {
        return html`...`;
    }
    if( this._isAudio) {
        return html`...`;
    }
}
```

media-player.js
```diff
@@ -353,7 +413,7 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 
 						<div id="d2l-labs-media-player-volume-level-container" class="${volumeLevelContainerClass}">
 							<div class="d2l-labs-media-player-rotated" id="d2l-labs-media-player-volume-level-background">
-								<d2l-seek-bar solid id="d2l-labs-media-player-volume-level" vertical="" value="${Math.round(this._volume * 100)}" aria-valuemin="0" aria-valuemax="100" aria-valuenow="${Math.floor(this._volume * 100)}" @position-change=${this._onPositionChangeVolume} @focus=${this._startUsingVolumeContainer} @focusout=${this._stopUsingVolumeContainer}></d2l-seek-bar>
+								<d2l-seek-bar solid id="d2l-labs-media-player-volume-level" vertical="true" value="${Math.round(this._volume * 100)}" aria-label="slider" aria-valuemin="0" aria-valuemax="100" aria-valuenow="${Math.floor(this._volume * 100)}" title="slider" @position-change=${this._onPositionChangeVolume} @focus=${this._startUsingVolumeContainer} @focusout=${this._stopUsingVolumeContainer}></d2l-seek-bar>
```
dlockhart, line 416: `vertical` is a [Boolean attribute](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes), so it should either be present or missing -- you don't need to set it to a value.

media-player.js
```diff
@@ -284,13 +311,25 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+			});
+		}
 		this._currentTime = 0;
+		this._determiningSourceType = true;
 		this._duration = 1;
 		this._hoveringVideoControls = false;
+		this._isAudio = false;
+		this._isVideo = false;
 		this._muted = false;
+		this._playbackSpeeds = ['0.25', '0.5', '0.75', '1.0', '1.25', '1.5', '2.0'];
```
dlockhart, line 332: If these are constant (don't change from player to player), they could just be defined outside of the class at the top of the file.

media-player.js
```diff
@@ -648,7 +797,19 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+	_determineSourceType() {
+		this._isAudio = false;
+		this._isVideo = false;
+		const video = document.createElement('video');
+		video.src = this.src;
+		video.addEventListener('canplay', () => {
+			this._determiningSourceType = false;
+			this._isAudio = video.videoHeight === 0;
+			this._isVideo = !this._isAudio;
+		});
```
dlockhart, line 812: This is assuming that if it's not a video, then the audio player can handle it. What if they pass in something else like an image or an audio format that can't be played?

media-player.js
```diff
@@ -36,8 +39,11 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 			loop: { type: Boolean },
 			poster: { type: String },
 			src: { type: String },
+			_audioBars: { type: Array, attribute: false },
 			_currentTime: { type: Number, attribute: false },
 			_duration: { type: Number, attribute: false },
+			_isAudio: { type: Boolean, attribute: false },
+			_isVideo: { type: Boolean, attribute: false },
```
dlockhart, line 46: Instead of 3 properties to control the state (`_isAudio`, `_isVideo` and `_determiningSourceType`), I'd suggest just having a single `_sourceType` property that either be "loading", "video" or "audio". That should simplify a lot of the logic elsewhere in the control. You could even have 4 states to handle the case where you've determined that you can't play the media file at all -- something like "unknown".

media-player.js
```diff
@@ -562,16 +633,94 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+			newAudioBars.push(newAudioBar);
+		});
+
+		this._audioBars = newAudioBars;
+
+		if (this._audioBarsOffset === this._audioBars.length - 1) this._audioBarsOffset = 0;
+		else this._audioBarsOffset += 1;
+	}
+
+	_getRgbOfAudioBar(i) {
```
dlockhart, line 682: It may be a good time to break some of this out into a separate "private" web component, like a `<d2l-labs-media-player-audio-bars>` component where you can centralize this logic. Because the way it is now, each time the audio bars change, your entire component is going to re-render. So every `50ms` the whole thing re-renders. I _think_ breaking it up into a separate component with its own lifecycle could help improve that, but I'm not totally sure -- @dbatiste any ideas here?

I'm also wondering if CSS transitions could be used instead of all the manual manipulation of colours.

You could also have sub-components for the media player buttons and a few other things that may make sense to factor out.

Jevan5: I also have an interval, `_startUpdatingCurrentTime()`, that runs as quickly as possible with a period of 0ms. This is to update the seek-bar with the current time of the media, whenever possible. `_startUpdatingCurrentTime()` updates the `_currentTime` property, and also causes re-rendering. Is this an issue too?

media-player.js
Jevan5: In the main `render()` function, use the function like this?:
```
render() {
    return html`
        <div id="d2l-labs-media-player-media-container">
            ...
            ${this._getContentAreaView()}
            ...
        </div>
    `;
}
```

media-player.js
Jevan5: `_determiningSourceType` is only used to determine whether or not to hide the media controls. What if there are only 3 states: `unknown`, `video`, and `audio`?

Then, `_determineSourceType()` can be:
```
this._sourceType = 'unknown';
const video = document.createElement('video');
video.src = this.src;
video.addEventListener('canplay', () => {
    this._sourceType = video.videoHeight === 0 ? 'audio' : 'video';
});
```

If the event fires a 'canplay' event, it is a media element. From there, just have to decide if it is audio or video. If the 'canplay' event is not fired, then the source is unsupported, and it remains 'unknown'.

media-player.js
dlockhart: Yep exactly!

media-player.js
dlockhart: LGTM

media-player.js
dlockhart: I'm pretty worried about the re-renders, because they're going to cause a reflow of the page and just generally cause slowness.

So I guess two things to try and improve:

1. `_startUpdatingCurrentTime()`: yeah running that every `0ms` isn't amazing. Ideally it would only be running a) if  media is actually playing and b) only as frequently as the granularity of the clock you need to update. So if the clock only updates every second, then no need to run it more often than that.
2. The audio colours: ideally this could happen using _only_ CSS animations/transitions, which would never cause a re-render. I can take a deeper look at what it's doing and see what's possible.

media-player.js
Jevan5: If we are always referencing the strings 'unknown', 'video', and 'audio', should these be referenced as string literals, or as defined variables?

```
// Literals

if (this._sourceType === 'unknown') { ... }
```
or
```
// Defined variables

const SOURCE_TYPES = { audio: 'audio', unknown: 'unknown', video: 'video' };

...

if (this._sourceType === SOURCE_TYPES.unknown) { ... }
```

dbatiste: Yeah we commonly factor those out in an enum like you've shown.

media-player.js
```diff
@@ -364,87 +424,98 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
-					<div id="d2l-labs-media-player-fullscreen" class="d2l-labs-media-player-control-element d2l-labs-media-player-control-element-last">
+					<div id="d2l-labs-media-player-fullscreen" class="d2l-labs-media-player-control-element d2l-labs-media-player-control-element-last" ?hidden="${!this._isVideo}">
 						<button ?hidden="${!FULLSCREEN_ENABLED}" class="d2l-labs-media-player-button" title="${fullscreenTooltip}" aria-label="${fullscreenTooltip}" @click=${this._toggleFullscreen}>
 							<d2l-icon class="d2l-labs-media-player-control-display" icon="${fullscreenIcon}"></d2l-icon>
 						</button>
 					</div>
 				</div>
 			</div>
+
+			<div id="d2l-labs-media-player-settings-menu-container" class="${this._usingSettingsMenu ? '' : 'hidden'}" style=${styleMap(settingsMenuContainerStyle)}>
```
dbatiste, line 441: I think it would be better to apply `hidden` attribute instead of a generic class name.  This is probably _relatively_ safe, but that's what we'd normally do.  If you switch it, be sure to update the CSS above as well.

media-player.js
```diff
@@ -364,87 +424,98 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+				<div class="d2l-labs-media-player-flex-filler"></div>
+				<d2l-menu id="d2l-labs-media-player-settings-menu">
```
dbatiste, line 443: Note: the root menu should be labelled.  `<d2l-menu label="...">`.  While not visible to sighted users, it gets announced to screen readers.

media-player.js
Jevan5: 1. I got feedback from designers that updating it every second was too slow. So, I reduced it to 250ms, but there was a bug with the seekbar (if you clicked somewhere on the seekbar, the knob would move there for a split-second, then move back) that was resolved when I reduced it to 0ms. Maybe I can look into a different way to resolve this issue, so that we can move back to 250ms.

2. There are 3 colours, `DAE4EB`, `A7D5F6`, and `8EB5D1`, represented in RGB. For simplicity, let's only consider the R portion, so `DA => 218`, `A7 => 167`, and `8E => 142`. During the first frame:
a) the left-most bar will have the value 218
b) the middle-left (25% along the row of bars) bar will have the value 167
c) the middle bar will have the value 142
d) the middle-right bar (75% along the row of bars) will have the value 167
e) the right-most bar will have the value ~218 (since it is almost exactly the left-most bar, in terms of the wrapping gradient)

On the second frame:
a) the left-most bar will have the value 218 + (167 - 218) / (number of bars / 4). This is calculated as having the same value as last time, except it is slowly becoming the value 167, determined by how many other bars are between it and the first quarter of the row
b) the middle-left bar will have the value 167 + (142 - 167) / (number of bars / 4)
c) ...
d) ...

The pattern sweeps those 3 colours across the bars. I tried looking into linear-gradients, but they don't wrap around. You can create a linear-gradient of `DAE4EB -> A7D5F6` between pixels `100 -> 200`, but I don't _think_ you can create a linear-gradient of `DAE4EB -> A7D5F6` between pixels `200 -> 250`, that finishes while wrapping from pixels `0 -> 50`, without breaking up the linear-gradient into sections, which I have done here.

media-player.js
Jevan5: I have css rules set for `#d2l-labs-media-player-settings-menu-container`, which set `display: flex;`. I think that is overriding the `hidden` attribute, and it does not work.

The CSS rule I use for this is pretty specific, so it I thought it would be okay:
```
#d2l-labs-media-player-settings-menu-container.hidden {
    display: none;
}
```

media-player.js
dbatiste: Yeah.  Breaking it out will at least help limit the work done when the life cycle runs, avoiding measurement or dom queries that aren't needed for such updates.  It will also help separate the concerns too, so good anyway imo.

Besides moving that logic into its own component, it would be interesting to explore options on how we can make a component that gets updated super efficiently, ex. on some interval.  Just spit-balling here... but it would be nice to utilize CSS properties that won't cause reflow, typical ones are `opacity` and `translate` see ([css triggers](https://csstriggers.com/)).  Maybe there is a directive we can use to update based on some observable, without doing a complete re-render. 🤔 

For the time, unfortunately updating text content is going to cause layout, but we can at least try to minimize the overhead.

media-player.js
dbatiste: So the risk is that there could be other CSS on the page that targets your element in legacy Edge.  The element having a generic class name increases the chances of it happening, which when using classes, we always put `d2l-component-name-*` in front of them.  In this case, the chance of it doing something different are pretty low, but I figured I would point it out since it's not something we'd do, and I think the style linter would catch it. If that other CSS was setting other properties, or was using `!important`, the specificity here wouldn't matter.  Not a huge deal, but definitely something to avoid until we are able to get rid of legacy Edge.

PR #14, opened 2020-10-09
Us118769/captions support

Jevan5:


src/fullscreen-api.js
```diff
@@ -0,0 +1,149 @@
+class FullscreenApi {
```
Jevan5, line 1: @dbatiste You mentioned implementing a simple third-party dependency that we were using, `screenfull`, a while ago. Does this belong here, or in `BrightspaceUI/core`?

media-player.js
```diff
@@ -263,35 +278,61 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 				justify-content: center;
 				width: 100%;
 			}
+
+			#d2l-labs-media-player-track-container {
+				align-items: center;
+				bottom: 3rem;
+				color: var(--d2l-color-white);
```
dlockhart, line 285: This isn't actually a color we support in our palette -- just use `#ffffff`.

media-player.js
```diff
@@ -263,35 +278,61 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+			}
+
+			#d2l-labs-media-player-track-container > div {
+				min-width: ${MIN_TRACK_WIDTH_PX}px;
+				width: 50%;
+			}
+
+			#d2l-labs-media-player-track-container > div > span {
+				background-color: rgba(0, 0, 0, 0.69);
+				font-family: 'Lato', sans-serif;
```
dlockhart, line 302: Hopefully you don't need to set the font to Lato here? You get Lato for free from the page.

media-player.js
```diff
@@ -559,6 +616,148 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 		this._videoClicked = !this._videoClicked;
 	}
 
+	async _onSlotChange(e) {
+		this._tracks = [];
+		const nodes = e.target.assignedNodes();
+		for (let i = 0; i < nodes.length; i++) {
+			const node = nodes[i];
+
+			if (node.nodeType === Node.ELEMENT_NODE && node.nodeName === 'TRACK') {
```
dlockhart, line 625: I'd inverse this `if`-statement and then `continue` to reduce the nesting.

```javascript
if (node.nodeType !== Node.ELEMENT_NODE || node.nodeName !== 'TRACK') continue;
```

media-player.js
```diff
@@ -559,6 +616,148 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+				clearInterval(interval);
+
+				resolve();
+			});
+		});
+
+		const oldTracks = this._media.querySelectorAll('track');
+		oldTracks.forEach(track => this._media.removeChild(track));
+
+		this._tracks.forEach(track => {
```
dlockhart, line 685: Could this rendering not happen inside `getMediaAreaView`? Like aren't these `<track>` elements supposed to go inside the audio/video elements?

media-player.js
```diff
@@ -559,6 +616,148 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+				trackElement.oncuechange = this._onCueChange.bind(this);
+				this._media.appendChild(trackElement);
+			}
+		});
+	}
+
+	_onPlaybackSpeedsMenuItemChange(e) {
+		this._media.playbackRate = e.target.value;
+
+		this.shadowRoot.getElementById('d2l-labs-media-player-playback-speeds').value = e.target.value;
```
dlockhart, line 710: Here and in the other `onMenuItemChange` handlers you're reaching into and modifying DOM nodes outside of the `render()` method. Really want to avoid that if you can.

README.md
```diff
@@ -103,6 +103,29 @@ this.document.querySelector('d2l-labs-media-player').addEventListener('loadeddat
 });
 ```
 
+## track
```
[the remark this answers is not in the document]
Jevan5: I noticed that the convention within `BrightspaceUI/core/components` was to label these headings as the name of the element/component. Still prefer to change the heading?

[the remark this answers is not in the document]
dlockhart: I personally would be up to you. In this case it's less about the element (which is just a native element) and more about how to add captions/subtitles to the media.

README.md
```diff
@@ -103,6 +103,29 @@ this.document.querySelector('d2l-labs-media-player').addEventListener('loadeddat
...
+<script type="module">
+    import '@brightspace-ui-labs/media-player/media-player.js';
+</script>
+<d2l-labs-media-player src="/video.webm">
+	<track src="/english-captions.srt" srclang="en" label="English" kind="captions">
+	<track src="/french-captions.vtt" srclang="fr" label="French" kind="captions">
+</d2l-labs-media-player>
+```
+
+**Properties**
```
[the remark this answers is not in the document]
Jevan5: It's a little different, because the attributes are required and only some of the kinds are supported. What if I left the specifications, but added a link to MDN as well?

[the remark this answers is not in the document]
dlockhart: Yep works for me!

media-player.js
Jevan5: When I open the element inspector, I see `html { ... --d2l-color-white: #fff; ... }`. I thought this was the css variable being set?

media-player.js
Jevan5: There is programmatic stuff that needs to be done:

If it is an SRT track, parse the file and set the cues of the track to be the captions/subtitles
`track.cues.forEach(cue => { ... });`

So, these tracks can't just be rendered as HTML within the `render()` method.

media-player.js
Jevan5: Again, the tracks have to be rendered programatically with JS methods. So, I can't change their properties in the `render()` method, because they don't exist there.

I can move some of the DOM modifications to the `render()` method, but not all of them.

PR #21, opened 2020-10-26
Us118769/ui changes

Jevan5:


media-player.js
```diff
@@ -307,6 +323,59 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+				height: 3rem;
+				justify-content: center;
+				margin: 0;
+				padding: 2px;
+				position: absolute;
+				width: 3rem;
+			}
+
+			#d2l-labs-media-player-audio-play-button:hover {
+				background: #CDD5DC;
```
dlockhart, line 342: All these colours are already in CSS variables that you can use:
https://github.com/BrightspaceUI/core/blob/master/components/colors/colors.js#L12

```
import '@brightspace-ui/core/components/colors/colors.js';

...

background-color: var(--d2l-color-mica);
```

media-player.js
```diff
@@ -409,18 +489,18 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+			<div class="${this._hidingCustomControls() ? 'hidden' : ''}" id="d2l-labs-media-player-media-controls" ?hidden="${NATIVE_CONTROLS}" style=${styleMap(mediaControlsStyle)} @mouseenter=${this._startHoveringControls} @mouseleave=${this._stopHoveringControls}>
 				<d2l-seek-bar fullWidth solid id="d2l-labs-media-player-seek-bar" value="${Math.floor(this.currentTime / this._duration * 100)}" aria-label="${this.localize('seekSlider')}" aria-orientation="horizontal" aria-valuemin="0" aria-valuemax="100" aria-valuenow="${Math.floor(this.currentTime / this._duration * 100)}" title="${this.localize('seekSlider')}" @drag-start=${this._onDragStartSeek} @drag-end=${this._onDragEndSeek} @position-change=${this._onPositionChangeSeek}></d2l-seek-bar>
 				<div id="d2l-labs-media-player-buttons">
-					<div class="d2l-labs-media-player-control-element">
-						<button class="d2l-labs-media-player-button" id="d2l-labs-media-player-play-button" title="${playTooltip}" aria-label="${playTooltip}" @click=${this._togglePlay}>
-							<d2l-icon class="d2l-labs-media-player-control-display" icon="${playIcon}"></d2l-icon>
+					<div class="d2l-labs-media-player-control-element ${this._sourceType}">
```
dlockhart, line 495: I'd use `classMap` for this -- which will let you use a namespaced CSS class (e.g. `d2l-labs-media-player-control-element-audio`) instead of just the generic `audio` which could collide with something else on the page.

Also, this source type CSS class could be placed much higher up in the DOM so that it only needs to go on one element. Then you can leverage the cascading nature of CSS to target elements further down.

media-player.js
```diff
@@ -386,20 +459,27 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 		const playTooltip = `${this._playing ? this.localize('pause') : this.localize('play')} (${KEY_BINDINGS.play})`;
 		const volumeTooltip = `${this._muted ? this.localize('unmute') : this.localize('mute')} (${KEY_BINDINGS.mute})`;
 
+		const controlDisplayStyle = { color: this._sourceType === SOURCE_TYPES.video ? '#ffffff' : '#494c4e' };
+		const mediaContainerStyle = { backgroundColor: this._sourceType === SOURCE_TYPES.audio ? '#ffffff' : '#000000', cursor: !this._hidingCustomControls() || this._sourceType === SOURCE_TYPES.unknown ? 'auto' : 'none', display: this._sourceType === SOURCE_TYPES.unknown ? 'none' : 'block' };
+		const mediaControlsStyle = { backgroundColor: this._sourceType === SOURCE_TYPES.video ? 'rgba(0, 0, 0, 0.69)' : '#ffffff' };
```
dlockhart, line 464: I think a clever use of CSS selectors could remove the need for all this. Something like:

```css
.d2l-labs-media-player-button.video d2l-icon {
    color: #ffffff;
}
.d2l-labs-media-player-button.audio d2l-icon {
    color: var(--d2l-color-ferrite);
}
```

media-player.js
```diff
@@ -875,33 +965,57 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+	}
+
+	_getErrorView() {
+		return this._message.type === MESSAGE_TYPES.error ? html`
+			<div id="d2l-labs-media-player-error">
+				<d2l-icon icon="tier1:alert"></d2l-icon>
+				<span>
+					${this._message.text}
+				</span>
+			</div>
```
dlockhart, line 994: Not sure what this is supposed to look like, but have you seen our `<d2l-alert>` component?

media-player.js
```diff
@@ -875,33 +965,57 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
 			case SOURCE_TYPES.audio:
 				return html`
-					<audio crossorigin="anonymous" id="d2l-labs-media-player-audio" ?controls="${NATIVE_CONTROLS}" ?autoplay="${this.autoplay}" ?loop="${this.loop}" preload="metadata" @ended=${this._onEnded} @loadeddata=${this._onLoadedData} @play=${this._onPlay} @pause=${this._onPause} @loadedmetadata=${this._onLoadedMetadata} @timeupdate=${this._onTimeUpdate} @volumechange=${this._onVolumeChange}>
+					<audio crossorigin="anonymous" id="d2l-labs-media-player-audio" ?controls="${NATIVE_CONTROLS}" ?autoplay="${this.autoplay}" ?loop="${this.loop}" preload="metadata" @ended=${this._onEnded} @error=${this._onError} @loadeddata=${this._onLoadedData} @play=${this._onPlay} @pause=${this._onPause} @loadedmetadata=${this._onLoadedMetadata} @timeupdate=${this._onTimeUpdate} @volumechange=${this._onVolumeChange}>
 						<source src="${this.src}"></source>
 					</audio>
 
 					<div id="d2l-labs-media-player-audio-bars-container">
+						<button id="d2l-labs-media-player-audio-play-button" @click=${this._togglePlay}>
+							<d2l-icon icon="${playIcon}"></d2l-icon>
```
dlockhart, line 1016: This button has no label -- at the very least it should have an `aria-label`.

.gitignore
```diff
@@ -1,5 +1,4 @@
 .vscode/
 node_modules/
-package-lock.json
```
dlockhart, line 3: Why are you including a lock file? Typically they're not useful for "library"-type repos like this. With a lock file, you're going to need to explicitly pick up new versions of things. And BSI is going to ignore it anyway.

PR #24, opened 2020-10-28
Add paused and ended attributes

AdmiralCoco:


media-player.js
```diff
@@ -50,7 +50,9 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 	static get properties() {
 		return {
 			autoplay: { type: Boolean },
+			ended: { type: Boolean },
```
Jevan5, line 53: Should these be in the list of `properties` that cause re-rendering of the component? It does not seem like they should. I think having the public `get ended()` and `get paused()` methods are sufficient.

AdmiralCoco: Good point, going to remove them here

Jevan5 (approved):

PR #27, opened 2020-10-29
Menu updates

dlockhart:
This is still incomplete, but I wanted to get something up so you can see where I'm going with this.

Still TODO:
- [x] Once button-icons support dark mode (in progress in core), switch from using translucent buttons to the dark theme
- [x] Make changes to the menu dark mode styles based on Jeff & Mike's updates (in progress in core) and adopt them here


media-player.js
```diff
@@ -512,17 +403,18 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+					<div id="d2l-labs-media-player-volume-container" @mouseenter="${this._startUsingVolumeContainer}" @mouseleave="${this._stopUsingVolumeContainer}">
+						<d2l-button-icon
+							class="d2l-dropdown-opener" 
+							icon="${volumeIcon}"
+							text="${volumeTooltip}"
+							theme="${ifDefined(theme)}"
+							?translucent="${translucent}"
+							@blur="${this._stopUsingVolumeContainer}"
+							@click="${this._toggleMute}"
+							@focus="${this._startUsingVolumeContainer}"></d2l-button-icon>
```
Jevan5, line 417: Q: Do you prefer to have attributes and events for elements on their own lines? I use word wrap in my IDE, and don't mind the long sequence of text. But I know that some people prefer to separate them for readability.

If you're used to separating them like this, I can adopt this practice as well. Consistency is 🔑

media-player.js
```diff
@@ -536,37 +428,27 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
 
-					<div class=${classMap(settingsMenuButtonContainerClass)}>
-						<button class="d2l-labs-media-player-button" id="d2l-labs-media-player-settings-button" label="${this.localize('settings')}" aria-label="${this.localize('settings')}" @click=${this._settingsButtonClick}>
-							<d2l-icon icon="tier1:gear"></d2l-icon>
-						</button>
-					</div>
+					<d2l-dropdown>
+						<d2l-button-icon class="d2l-dropdown-opener" icon="tier1:gear" text="${this.localize('settings')}" ?translucent=${translucent} theme="${ifDefined(theme)}"></d2l-button-icon>
+						<d2l-dropdown-menu id="d2l-labs-media-player-settings-menu" no-pointer theme="${ifDefined(theme)}" vertical-offset="10">
+							<d2l-menu label="${this.localize('settings')}" theme="${ifDefined(theme)}">
```
Jevan5, line 434: I'm not sure where this positions the menu, but I know that we want it tucked into the bottom-right corner of the player. Bottom of the menu would be resting on the scrub bar, and the right side of the menu would be resting along the right side of the player.

Jevan5:
I think `<button id="d2l-labs-media-player-audio-play-button" ...><d2l-icon ...></d2l-icon></button>` should be replaced with `<d2l-button-icon ...></d2l-button-icon>` as well.

media-player.js
dlockhart: Mostly I start wrapping things when the line starts to get long... typically around 80 characters so that when you view the code in places like on GitHub it's readable.

media-player.js
dlockhart: Ok we can tweak this as dropdown content (`d2l-dropdown-menu` in this case) allows you to provide a boundary object.

dlockhart:
I've switched this over to use the new dark theme for button-icons -- should be good to review and merge whenever you like!

Jevan5 (approved):

PR #29, opened 2020-11-03
Jevan5

Jevan5:


Jevan5:
- Updated the gradient algorithm using the method described [here](https://stackoverflow.com/questions/22607043/color-gradient-algorithm)
Audio Bars

- Allow for three lines of subtitles to be viewable before blocking the play button
Screen Shot 2020-11-02 at 11 58 39 PM

- Provide the option to allow for media downloads. Controlled with `allow-download` attribute. Also, tuck the settings menu into the bottom-right corner in LTR and RTL mode
Screen Shot 2020-11-03 at 12 00 39 AM

media-player.js
```diff
@@ -742,6 +783,17 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+		const link = document.createElement('a');
+		link.download = 'download';
+		link.target = '_blank';
+		const url = new Url(this.src);
+		const searchParams = url.searchParams;
+		searchParams.append('download', 'true');
+		link.href = url.toString();
+		link.click();
+	}
+
```
Jevan5, line 796: This was a last minute requirement for Lessons, and only opens the video in a new window. This will be improved in another iteration.

media-player-audio-bars.js
```diff
@@ -1,24 +1,78 @@
...
-			playing: { type: Boolean }
+			playing: { type: Boolean },
+			_visibleAudioBars: { type: Array, attribute: false },
 		};
 	}
 
 	static get styles() {
 		return css`
+			:host {
+				width: 100%;
```
dlockhart, line 70: I think `display: block` is likely more what you were looking for? Custom elements are inline by default.

media-player.js
```diff
@@ -82,14 +84,14 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
-				min-height: 11rem;
+				min-height: 15rem;
 				overflow: hidden;
 				position: relative;
 				width: 100%;
 			}
 
 			.d2l-labs-media-player-type-is-audio {
-				background-color: #ffffff;
+				background-col: #ffffff;
```
dlockhart, line 94: `background-col` isn't a thing... `background-color` is. I'd recommend [enabling style linting](https://blog.d2l.dev/2020/08/04/style-linting/) so it catches things like this for you.

media-player.js
dlockhart: I'd recommend `<d2l-menu-item-link href="${linkHref}" target="_blank">`.

media-player.js
Jevan5: This is a typo, no idea how I managed to do this

media-player-audio-bars.js
Jevan5: Using `display: block;`, the `<d2l-labs-media-player-audio-bars>` element does not fit its parent and has a width of 0

media-player.js
Jevan5: Should clicking a `<d2l-menu-item-link>` close the dropdown menu that it is in? It is opening the link in a new tab, but the menu on the original page remains open

dlockhart: Normally these don't open in a new window, so clicking them would cause a full page navigation. So maybe you've found a defect where if a target is set, we should be closing the menu. @dbatiste any thoughts here?

dbatiste: Yeah, I think that was the original thinking.  Seems reasonable to close the menu.  Should we only close it in the `_blank` case?

media-player-audio-bars.js
```diff
@@ -67,79 +152,121 @@ class MediaPlayerAudioBars extends LitElement {
...
 	}
 
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		new ResizeObserver((entries) => {
+			for (const entry of entries) {
+				const { width } = entry.contentRect;
+
+				const fontSize = getComputedStyle(document.documentElement).fontSize;
```
dbatiste, line 162: Is it necessary to do this computed style work in the resize observer, or even in this component?  I think it would be fine to get the root font size once, outside of this component so that it's only ever done once no matter how many players on on the page.  The root font size should never change after the page is loaded.  I would try to minimize the amount of work that is done in such a callback.

media-player.js
dlockhart: We could, or just always close it for simplicity and for snappiness since people would expect it to close either way?

dbatiste: Do we need to have the `download` property added to `d2l-menu-item-link` and passed down to the underlying anchor?

media-player.js
```diff
@@ -742,6 +783,17 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 		if (this._sourceType === SOURCE_TYPES.video) this._onVideoClick();
 	}
 
+	_onDownloadClick() {
+		const link = document.createElement('a');
+		link.download = 'download';
+		link.target = '_blank';
+		const url = new Url(this.src);
+		const searchParams = url.searchParams;
+		searchParams.append('download', 'true');
```
MichaelKovalchuk, line 792: can we change this to ('attachment', 'true`)?

media-player.js
```diff
@@ -247,21 +263,26 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
 				color: white;
 				box-shadow: 0.3rem 0 0 rgba(0, 0, 0, 0.69), -0.3rem 0 0 rgba(0, 0, 0, 0.69);
+				line-height: 1.35rem;
 				white-space: pre-wrap;
 			}
 
+			#d2l-labs-media-player-audio-play-button-container {
+				background-color: white;
+				border-radius: 9px;
+				padding: 0px;
```
dbatiste, line 273: Nit: zero value dimensions should be unitless.  I think if the style linter is added, it will catch this stuff for us.

Jevan5:
I wiped my `node_modules` and `package-lock.json` locally, and am using the same versions of node and npm as the travis-ci build. The linting is passing fine on my local instance, but travis-ci is complaining about failed tests. The linting version in travis-ci expects static methods to come before everything, but that is not the case with my local instance.

Jevan5:
> I wiped my `node_modules` and `package-lock.json` locally, and am using the same versions of node and npm as the travis-ci build. The linting is passing fine on my local instance, but travis-ci is complaining about failed tests. The linting version in travis-ci expects static methods to come before everything, but that is not the case with my local instance.

May have been related to travis-ci  maintaining older versions of `node_modules`. With @dbatiste's recommendation, adding `cache: npm: false` to `.travis.yml` resolved the issue.

package.json
```diff
@@ -48,10 +50,12 @@
   },
   "dependencies": {
     "@brightspace-ui/core": "^1",
+    "@brightspace-ui/stylelint-config": "^0",
```
dbatiste, line 53: This dep, and also the stylelint dep should be dev dependencies.

Jevan5: Absolutely. The guy who wrote this [blog](https://blog.d2l.dev/2020/08/04/style-linting/) is to blame:
```
npm i @brightspace-ui/stylelint-config
npm i stylelint
```

dbatiste: 😬

media-player.js
```diff
@@ -520,112 +547,302 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
-		this._tracks = [];
-		const nodes = e.target.assignedNodes();
-		for (let i = 0; i < nodes.length; i++) {
-			const node = nodes[i];
+	_getErrorAlertView() {
+		return this._message.type === MESSAGE_TYPES.error ? html`
+			<div class="d2l-labs-media-player-full-area-centred">
+				<d2l-alert button-text="${this.localize('retry')}" type="critical" @d2l-alert-button-press=${this._onAlertButtonPress}>
+					<div id="d2l-labs-media-player-alert-inner">
+						<img src="../assets/img/media-play-broken.svg">
```
dbatiste, line 635: Is this working as expected when used in the LMS, with BSI coming from the CDN?   The reason I ask is... in such a case, I think the browser will resolve this path relative to the the LMS domain, not the CDN where the asset is located, resulting in a 404 for the image.

Assuming this is an issue that needs to be addressed, I think there are options:

1. Place the SVG inline here (advantage of saving an HTTP request, at the cost of bloating this HTML for something that's hopefully rarely needed).

2. Resolving the path manually... `import.meta.url.substring(0, import.meta.url.lastIndexOf('/'))`. I don't think we have a lot of experience with this API, but I am using it in `d2l-htmleditor`, so far successfully.

3. Import this from a JS module and write out the SVG using the `unsafeHTML` directive. This is the approach `d2l-icon` uses because it requires loading icons on demand, with some runtime processing of the SVG.

media-player.js
```diff
@@ -520,112 +547,302 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
+
+						<span style="">
```
dbatiste, line 637: Nit: empty style attribute.

media-player.js
Jevan5: For #2, this works locally and in LE:
```
<img src="${`${import.meta.url.substring(0, import.meta.url.lastIndexOf('/'))}/assets/img/media-play-broken.svg`}">
```

Is that what you had in mind?

dbatiste: Yeah that's basically what I had in mind.  The same thing is basically being done [here](https://github.com/BrightspaceUI/htmleditor/blob/master/htmleditor.js#L66).  We _could_ factor this into a helper... maybe not worth it though.

media-player.js
```diff
@@ -520,112 +547,302 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
...
-		this._tracks = [];
-		const nodes = e.target.assignedNodes();
-		for (let i = 0; i < nodes.length; i++) {
-			const node = nodes[i];
+	_getErrorAlertView() {
+		return this._message.type === MESSAGE_TYPES.error ? html`
+			<div class="d2l-labs-media-player-full-area-centred">
+				<d2l-alert button-text="${this.localize('retry')}" type="critical" @d2l-alert-button-press=${this._onAlertButtonPress}>
+					<div id="d2l-labs-media-player-alert-inner">
+						<img src="${`${import.meta.url.substring(0, import.meta.url.lastIndexOf('/'))}/assets/img/media-play-broken.svg`}">
```
dlockhart, line 635: How is BSI going to know to copy the file though? This might work in a dev build of BSI, but I think the Rollup build won't know where to find the file in a production build.

The inline SVG option Dave mentioned is the easiest way around this, and you can combine it with `<d2l-icon-custom>` for built-in sizing & colouring:
https://github.com/BrightspaceUI/core/tree/master/components/icons#custom-svg-icons

If you stick with an `img` tag, you need `alt=""` to mark it as decorative.

PR #32, opened 2020-11-09
Handle relative URLs

Jevan5:
Media URLs in the Legacy Content Tool are relative.

Ex.
```
/content/course1/sample-video.mp4?d2lSessionVal=IkuWvsshZrwnCEfy5bmvjv346&ou=6609
```

When constructing a [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) for the download button, you cannot provide only a relative URL string.

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -613,10 +613,17 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 		return str.join('');
 	}
 
+	_getAbsoluteUrl(url) {
+		const a = document.createElement('a');
+		a.setAttribute('href', url);
+
+		return a.href;
+	}
+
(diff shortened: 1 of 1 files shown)


media-player.js
```diff
@@ -613,10 +613,17 @@ class MediaPlayer extends InternalLocalizeMixin(RtlMixin(LitElement)) {
 		return str.join('');
 	}
 
+	_getAbsoluteUrl(url) {
+		const a = document.createElement('a');
+		a.setAttribute('href', url);
+
+		return a.href;
```
Jevan5, line 620: If you set the `href` attribute of an anchor element, the `href` property resolves to the absolute URL.

rschick (approved):

PR #35, opened 2020-11-10
Fix Safari Not Loading Player

Jevan5:


media-player.js
```diff
@@ -578,14 +578,6 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
 		};
 		this._sourceType = SOURCE_TYPES.unknown;
-		const video = document.createElement('video');
-		video.src = this.src;
-
-		video.addEventListener('canplay', () => {
-			this._sourceType = video.videoHeight === 0 ? SOURCE_TYPES.audio : SOURCE_TYPES.video;
-		});
-
-		video.addEventListener('error', this._onError.bind(this));
```
Jevan5, line 588: This is the method for determining the source type. I think the issue with Safari had something to do with rendering the same video twice.

media-player.js
```diff
@@ -684,20 +676,30 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+		const mediaStyle = {};
+		if (this._sourceType === SOURCE_TYPES.unknown || this._loading) {
+			mediaStyle.display = 'none';
+		}
 
 		switch (this._sourceType) {
-			case SOURCE_TYPES.video:
+			case SOURCE_TYPES.unknown:
+				this._determineSourceType();
+			case SOURCE_TYPES.video: // eslint-disable-line no-fallthrough
```
Jevan5, line 691: Rendering a `<video>` in two cases:
- The source type is not known: Render a hidden video, and check what the boundaries are of the element once it has loaded
- The source type is a video: Render a visible video

media-player.js
```diff
@@ -814,6 +818,16 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
 	}
 
+	_onDurationChange(e) {
+		if (this._sourceType === SOURCE_TYPES.unknown) {
+			this._sourceType = this.shadowRoot.getElementById('d2l-labs-media-player-video').videoHeight === 0 ? SOURCE_TYPES.audio : SOURCE_TYPES.video;
+
+			if (this._sourceType === SOURCE_TYPES.audio) return;
+		}
+
+		this._duration = e.target.duration;
```
Jevan5, line 828: Set the duration earlier in the loading process. Faster than waiting for `loadedmetadata`

test/media-player.visual-diff.js
```diff
@@ -24,7 +24,6 @@ describe('d2l-labs-media-player', () => {
 	[
 		'video',
 		'audio',
-		'error'
```
Jevan5, line 27: Visual diff for error has to be removed for now. Since the loading spinner is displayed momentarily, I will have to figure out how to wait for the loading spinner and alert drop to settle before taking a screenshot.

media-player.js
```diff
@@ -714,16 +706,18 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 				return html`
 					<audio
```
Jevan5, line 707: After the video element has loaded and the media turns out to be audio, this audio element will render instead of the video element.

media-player.js
```diff
@@ -828,13 +841,36 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
 
-	_onLoadedMetadata($event) {
+	_onLoadedDataAudio() {
+		if (this._sourceType === SOURCE_TYPES.audio) {
+			this._onLoadedData();
+		}
+	}
+
+	_onLoadedDataVideo() {
+		if (this._sourceType === SOURCE_TYPES.video) {
```
Jevan5, line 851: Only fire the event if the source type is accurate. If the video element loads and the media type is audio, this function will still be called even though the audio element has not finished loading yet.

media-player.js
```diff
@@ -814,6 +819,14 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		this._dragging = true;
 	}
 
+	_onDurationChange(e) {
+		if (this._sourceType === SOURCE_TYPES.unknown) {
+			this._sourceType = this.shadowRoot.getElementById('d2l-labs-media-player-video').videoHeight === 0 ? SOURCE_TYPES.audio : SOURCE_TYPES.video;
+		}
```
AdmiralCoco, line 825: This sets the sourceType, based on `videoType` (if the sourceType is unknown?)

AdmiralCoco (approved):

media-player.js
Jevan5: `this._sourceType` is originally `"unknown"`, because the media has not loaded. Once it has loaded, it is still unknown until we perform this check. Audio will load without height, while video will load with height.

Once `this._sourceType` is `"video"` or `"audio"`, we can render the appropriate element.

PR #39, opened 2020-11-10
Center Audio Play Button

Jevan5:


media-player-audio-bars.js
```diff
@@ -46,7 +46,7 @@ AUDIO_BARS_GRADIENTS.forEach((gradient, i) => {
 		gradient.to = AUDIO_BARS_GRADIENTS[0].from;
 	}
 });
-const AUDIO_BAR_HEIGHTS = [35, 54, 65, 86, 81, 67, 100, 90, 100, 98, 75, 99, 98, 96, 96, 95, 65, 50, 50, 60, 64, 54, 50, 44, 45, 46, 47, 45, 53, 67, 68, 76, 65, 63, 67, 91, 97, 86, 88, 86, 84, 84, 82, 0, 0, 0, 0, 0, 0, 0, 0, 0, 69, 70, 68, 67, 72, 76, 86, 81, 83, 67, 65, 67, 21, 44, 45, 97, 86, 85, 81, 84, 70, 65, 67, 67, 72, 79, 76, 76, 63, 44, 42, 50, 54, 49, 33, 42, 44, 33, 38, 38, 36, 37, 35, 34];
+const AUDIO_BAR_HEIGHTS = [35, 54, 65, 86, 81, 67, 100, 90, 100, 98, 75, 99, 98, 96, 96, 95, 65, 50, 50, 60, 64, 54, 50, 44, 45, 46, 47, 45, 53, 67, 68, 76, 65, 63, 67, 91, 97, 86, 88, 86, 84, 84, 82, 0, 0, 0, 0, 0, 0, 0, 0, 69, 70, 68, 67, 72, 76, 86, 81, 83, 67, 65, 67, 21, 44, 45, 97, 86, 85, 81, 84, 70, 65, 67, 67, 72, 79, 76, 76, 63, 44, 42, 50, 54, 49, 33, 42, 44, 33, 38, 38, 36, 37, 35, 34];
```
Jevan5, line 49: The number of displayed audio bars adjusts to the dynamic size of the player. If the size is too small for all the bars, it selects the ones from the middle. If the size is large enough for the bars, it displays all bars in the centre of the player.

It was centring properly in cases where the player was too small. When the player was large enough, it was displaying all the bars which were not centred. The reason was that the number of bars is even, while there should be an odd number of bars hidden behind the button, and equal amounts on both sides.

Made the number of audio bars odd, so that centring takes place normally.

rschick (approved):

PR #40, opened 2020-11-11
Retry load on error

Jevan5:
Fix for https://trello.com/c/KeAFPxhi/163-media-player-chrome-nle-content-when-video-file-request-fails-loading-spinner-continues-indefinitely-instead-of-displaying-retry

I'm not sure why this stopped working. I had tested the previous version of the component just a few days ago, and it was working. The `error` event is no longer being emitted by the `<video>`, and I found out that it is the `<source>` element that is really emitting the event.

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -668,8 +668,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	}
 
 	_getMediaAreaView() {
-		if (!this.src) return null;
-		if (this._message.type === MESSAGE_TYPES.error) return null;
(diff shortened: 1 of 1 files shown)


MarquisLP:
Yeah, it's strange how it suddenly stopped working. During testing yesterday morning, I could have sworn the Retry button was working in Content and NLE, but I'm not 100% sure. It definitely stopped working in the afternoon, however.

From what I've skimmed online, it seems like whether the `error` event is emitted by `<source>` or `<video>` depends on browser and version. But it seems unlikely that our browsers would have had significant updates in the span of just a few days/hours...

MarquisLP (approved):

PR #41, opened 2020-11-11
Fix captions cut-off on mobile

MarquisLP:


Jevan5 (changes requested):
On browsers, the width should be 50% of the width of the container, with a minimum of 250px.

In mobile, the width could be 100% of the width of the container, to compensate for smaller screens.

This has to be done conditionally on the user agent, because this change is currently breaking the design by affecting browsers as well.

rschick:
Please use CSS media selector based on screen size and not user agent detection.

MarquisLP:
I set the minimum width of the CSS selector to `767px` to account for landscape mode as well. But we can tweak that number if needed.

Jevan5 (approved):
Worked in my local instance

PR #43, opened 2020-11-12
Update Langterms

Jevan5:
Add fallback terms from `en.js` when a language is missing a lang-term.


rschick (approved):
Should we update en.js to use double quotes around object keys too?

Jevan5:
> Should we update en.js to use double quotes around object keys too?

Serge will populate the files without double quotes again. The double quotes are only because of the script.

If inconsistency is a problem, we can change it. Other than that, this poses no issues.

PR #47, opened 2020-11-12
Add iOS exceptions for rendering native controls

AdmiralCoco:
- Hide custom controls on iOS
- Add min-height CSS case for media-container
- Prevent togglePlay from running (let native controls handle play toggling)

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -23,7 +23,6 @@ import ResizeObserver from 'resize-observer-polyfill';
 import { RtlMixin } from '@brightspace-ui/core/mixins/rtl-mixin';
 import { styleMap } from 'lit-html/directives/style-map';
 
-const IS_IOS = /iPad|iPhone|iPod/.test(navigator.platform);
 const FULLSCREEN_ENABLED = fullscreenApi.isEnabled;
 const HIDE_DELAY_MS = 3000;
 const KEY_BINDINGS = {
@@ -36,7 +35,7 @@ const MESSAGE_TYPES = {
 	success: 2
 };
 const MIN_TRACK_WIDTH_PX = 250;
-const NATIVE_CONTROLS = !document.createElement('video').canPlayType;
+const IS_IOS = /iPad|iPhone|iPod/.test(navigator.platform);
 const PLAYBACK_SPEEDS = ['0.25', '0.5', '0.75', '1.0', '1.25', '1.5', '2.0'];
 const SEEK_BAR_UPDATE_PERIOD_MS = 0;
 const SOURCE_TYPES = {
@@ -89,7 +88,6 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 			}
 
 			#d2l-labs-media-player-media-container {
-				min-height: 15rem;
 				overflow: hidden;
 				position: relative;
 				width: 100%;
@@ -414,7 +412,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		const playTooltip = `${this._playing ? this.localize('pause') : this.localize('play')} (${KEY_BINDINGS.play})`;
 		const volumeTooltip = `${this._muted ? this.localize('unmute') : this.localize('mute')} (${KEY_BINDINGS.mute})`;
 
-		const mediaContainerStyle = { cursor: !this._hidingCustomControls() || this._sourceType === SOURCE_TYPES.unknown ? 'auto' : 'none' };
+		const mediaContainerStyle = {
+			cursor: !this._hidingCustomControls() || this._sourceType === SOURCE_TYPES.unknown ? 'auto' : 'none',
+			minHeight: this.isIOSVideo ? 'auto' : '15rem'
+		};
 		const trackContainerStyle = { bottom: this._hidingCustomControls() ? '3px' : 'calc(1.8rem + 20px)' };
 		const trackSpanStyle = { fontSize: `${this._trackFontSizeRem}rem`, lineHeight: `${this._trackFontSizeRem * 1.2}rem` };
 
@@ -444,14 +445,14 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		<div id="d2l-labs-media-player-media-container" class=${classMap(mediaContainerClass)} style=${styleMap(mediaContainerStyle)} @mousemove=${this._onVideoContainerMouseMove} @keydown=${this._listenForKeyboard}>
 			${this._getMediaAreaView()}
 
-			${IS_IOS ? null : html`
+			${this.isIOSVideo ? null : html`
 			<div id="d2l-labs-media-player-track-container" style=${styleMap(trackContainerStyle)} @click=${this._onTrackContainerClick}>
 				<div>
 					<span style=${styleMap(trackSpanStyle)} role="status">${this._trackText}</span>
 				</div>
 			</div>
 
-			<div class=${classMap(mediaControlsClass)} id="d2l-labs-media-player-media-controls" ?hidden="${NATIVE_CONTROLS}" @mouseenter=${this._startHoveringControls} @mouseleave=${this._stopHoveringControls}>
+			<div class=${classMap(mediaControlsClass)} id="d2l-labs-media-player-media-controls" @mouseenter=${this._startHoveringControls} @mouseleave=${this._stopHoveringControls}>
 				<d2l-seek-bar
 					id="d2l-labs-media-player-seek-bar"
 					fullWidth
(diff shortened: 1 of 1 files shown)


AdmiralCoco:
image
Captions can be selected with the native controls (only in fullscreen)

media-player.js
```diff
@@ -698,7 +700,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 					<video
 						id="d2l-labs-media-player-video"
 						style=${styleMap(mediaStyle)}
-						?controls="${NATIVE_CONTROLS}"
+						?controls="${IS_IOS || NATIVE_CONTROLS}"
```
rschick, line 703: Could we move this logic to where we calculate NATIVE_CONTROLS, instead of having it in the render method?

media-player.js
```diff
@@ -1048,7 +1050,11 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	}
 
 	_onVideoClick() {
-		this._togglePlay();
+		// Given that we are currently not rendering custom controls on iOS, we let the native
+		// controls/player handle the play/pause toggling
+		if (!IS_IOS) {
```
rschick, line 1055: Should this check "NATIVE_CONTROLS"?

media-player.js
```diff
@@ -1048,7 +1050,11 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+			this._togglePlay();
+		}
 		this._showControls(true);
```
rschick, line 1058: Should this be conditional as well, or is there logic in _showControls to check if we're using native controls?

rschick (approved):

Jevan5:
I recently had a look at `const NATIVE_CONTROLS = !document.createElement('video').canPlayType`, and this call does nothing. I got it from a custom video player tutorial I was looking at a while ago, and I think it is outdated. `canPlayType` is [supported](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType) by all modern desktop and mobile browsers. Therefor, `NATIVE_CONTROLS` is always false, and pointless.

I think we should replace `const NATIVE_CONTROLS = !document.createElement('video').canPlayType` with `const NATIVE_CONTROLS = /iPad|iPhone|iPod/.test(navigator.platform)`, since this is the only case we use native controls.

media-player.js
AdmiralCoco: With native controls it looks like it's always displayed - I think some of (or maybe the rest) of the code in this block could potentially be skipped as well. Will test again

Jevan5:
`#d2l-labs-media-player-media-container` has its min-height set explicitly in the `get styles()` function. This is redundant and should be removed.

Alternative: Not sure if it is good practise, but since the `IS_IOS` check it outside the class definition, you could write this conditional in the `get styles()` function. `get styles()` is only called once, while you are continuously checking this conditional within `render()`.

media-player.js
Jevan5: I think you probably want to skip the rest of the code as well, because the remaining code handles custom controls and double-clicking for fullscreen (which we can't control through Javascript)

Jevan5: I think you will want to skip the remainder of the code as well. It is for showing the custom controls temporarily, and to handle double-clicking to request fullscreen

media-player.js
```diff
@@ -415,7 +414,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		const playTooltip = `${this._playing ? this.localize('pause') : this.localize('play')} (${KEY_BINDINGS.play})`;
 		const volumeTooltip = `${this._muted ? this.localize('unmute') : this.localize('mute')} (${KEY_BINDINGS.mute})`;
 
-		const mediaContainerStyle = { cursor: !this._hidingCustomControls() || this._sourceType === SOURCE_TYPES.unknown ? 'auto' : 'none' };
+		const mediaContainerStyle = {
+			cursor: !this._hidingCustomControls() || this._sourceType === SOURCE_TYPES.unknown ? 'auto' : 'none',
+			...(this.NATIVE_CONTROLS || this._sourceType ===  SOURCE_TYPES.audio) && { minHeight: '15rem' },
```
Jevan5, line 419: Nit: Extra space between `===` and `SOURCE_TYPES.audio`

AdmiralCoco:
>Alternative: Not sure if it is good practise, but since the IS_IOS check it outside the class definition, you could write this conditional in the get styles() function. get styles() is only called once, while you are continuously checking this conditional within render().

It looks like you can't have any values besides other `css` strings, or numbers [link](https://lit-element.polymer-project.org/guide/styles#expressions) -  we could use `unsafeCSS` but it's only here, so I just kept it as is (and removing it from the `styles()` block

media-player.js
```diff
@@ -444,13 +446,14 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		<div id="d2l-labs-media-player-media-container" class=${classMap(mediaContainerClass)} style=${styleMap(mediaContainerStyle)} @mousemove=${this._onVideoContainerMouseMove} @keydown=${this._listenForKeyboard}>
 			${this._getMediaAreaView()}
 
+			${NATIVE_CONTROLS && this._sourceType === SOURCE_TYPES.video ? null : html`
```
Jevan5, line 449: If the client is iOS and they are listening to an audio file, our controls are shown, except for the volume control.

Are the captions behaving?

AdmiralCoco: > If the client is iOS and they are listening to an audio file, our controls are shown, except for the volume control.
Yep that's correct - I was thinking of reducing the checks by making a NATIVE_CONTROLS getter (which also checked for if it was a video), but I couldn't decide if that was more or less clear

>Are the captions behaving?
Yes, it shows the player captions correctly

media-player.js
```diff
@@ -443,13 +445,14 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
 			<div id="d2l-labs-media-player-track-container" style=${styleMap(trackContainerStyle)} @click=${this._onTrackContainerClick}>
 				<div>
 					<span style=${styleMap(trackSpanStyle)} role="status">${this._trackText}</span>
 				</div>
 			</div>
 
-			<div class=${classMap(mediaControlsClass)} id="d2l-labs-media-player-media-controls" ?hidden="${NATIVE_CONTROLS}" @mouseenter=${this._startHoveringControls} @mouseleave=${this._stopHoveringControls}>
+			<div class=${classMap(mediaControlsClass)} id="d2l-labs-media-player-media-controls" @mouseenter=${this._startHoveringControls} @mouseleave=${this._stopHoveringControls}>
```
AdmiralCoco, line 455: Removed the `hidden` check, as we need to render the seekbar for audio iOS

media-player.js
```diff
@@ -468,7 +471,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 				<div id="d2l-labs-media-player-buttons">
 					<d2l-button-icon icon="${playIcon}" text="${playTooltip}"  @click="${this._togglePlay}" theme="${ifDefined(theme)}"></d2l-button-icon>
 
-					<div id="d2l-labs-media-player-volume-container" @mouseenter="${this._startUsingVolumeContainer}" @mouseleave="${this._stopUsingVolumeContainer}">
+					<div id="d2l-labs-media-player-volume-container" @mouseenter="${this._startUsingVolumeContainer}" @mouseleave="${this._stopUsingVolumeContainer}" ?hidden="${NATIVE_CONTROLS}">
```
AdmiralCoco, line 474: Volume can't be controlled on iOS, so it's hidden here

media-player.js
Jevan5: If the iOS check does not mean that we are using native controls in the case of audio, I think the initial check should be kept as `const IS_IOS = ...`, and then have a getter which addresses the audio check as well

Jevan5 (approved):

media-player.js
```diff
@@ -379,6 +378,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		this._media.currentTime = time;
 	}
 
+	get isIOSVideo() {
+		return IS_IOS && this._sourceType === SOURCE_TYPES.video;
+	}
```
AdmiralCoco, line 383: @Jevan5 This is sort of what I was thinking but not sure if it's necessary..I don't feel strongly about having it
(not sure if it is what you had in mind though)

Jevan5: That is what I had in mind. If we just kept the `const NATIVE_CONTROLS`, it wouldn't fully control what its name suggests

PR #48, opened 2020-11-12
No target blank for download link

MichaelKovalchuk:


MichaelKovalchuk:
I've tested this in all non mobile browsers. since we have Content-Disposition, we no longer need target blank. 
The exception may be iOS, but since we show native controls there - this may not be a problem

Jevan5 (approved):

PR #52, opened 2020-11-13
It will stop flicker loop by flooring the height

MichaelKovalchuk:


MichaelKovalchuk:
The height is a decimal number. if i make it integer - the flickering does not happen. the difference can be as small as 0.001px

Jevan5 (approved):

PR #56, opened 2020-12-01
Delay Load with Poster

Jevan5:


media-player.js
```diff
@@ -732,7 +732,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 						?autoplay="${this.autoplay}"
 						?loop="${this.loop}"
 						poster="${ifDefined(this.poster)}"
-						preload="auto"
+						preload="${this.poster ? 'metadata' : 'auto'}"
```
Jevan5, line 735: Two of the available values for `preload` are `metadata` and `auto`. `metadata` loads less data upfront, and only loads more data as the video plays. `auto` loads more data upfront. The only reason that `auto` was being used was because Safari does not display the first frame of the video if `preload="metadata"` is set. In the case where a `poster` is provided, there is no need to use `auto` anymore.

rschick (approved):

PR #61, opened 2020-12-03
Add crossorigin to README

Jevan5:


Jevan5:
This should have been in the last PR. It could have waited until the next significant change, but I need to bump the version in NPM as well. No longer using `frau-ci`, so will need to start using `fix:`, `feat:`, and `perf:` to describe version bump.

rschick (approved):

PR #64, opened 2020-12-09
Lang double quotes

dlockhart:
Switching the English translations to use quoted keys and double-quotes for values so that Serge can properly encode things.

Also applying the translations from #58 as the above issue caused them to become corrupted. Should work fine next time around.


Jevan5 (approved):

lang/ar.js
```diff
@@ -1,23 +1,23 @@
 /* eslint quotes: 0 */
 export default {
-	"captions": "Captions",
+	"captions": "التسميات التوضيحية",
+	"closedCaptionsAcronym": "CC",
```
dlockhart, line 4: I added this entry into all of them as it was a new term that was missing.

lang/ar.js
```diff
@@ -1,23 +1,23 @@
...
 	"off": "إيقاف التشغيل",
 	"pause": "إيقاف مؤقت",
 	"play": "تشغيل",
 	"playbackSpeed": "سرعة التشغيل",
-	"retry": "Retry",
+	"retry": "إعادة المحاولة",
 	"seekSlider": "شريط تمرير التقديم/الترجيع",
 	"settings": "settings",
-	"speed": "speed",
-	"subtitles": "الترجمة",
```
dlockhart, line 20: I removed `subtitles` as it was removed from the English translations.

lang/en.js
```diff
@@ -1,22 +1,23 @@
+/* eslint quotes: 0 */
```
dlockhart, line 1: This file was the source of most of the problems. Make sure you keep the double-quotes around both the keys and the values.

PR #72, opened 2021-02-23
feat: add an optional attribute download-src to download from.

qwangd2l:


media-player.js
```diff
@@ -673,14 +688,23 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	_getDownloadButtonView() {
 		if (!this.allowDownload) return null;
 
-		const linkHref = this._getDownloadLink();
+		if (this.downloadSrc) {
+			return html`
+				<d2l-menu-item-link @click=${this._downloadFromSrc} text="${this.localize('download')}"></d2l-menu-item-link>
```
Jevan5, line 693: If this is no longer a link (no `href`), does this still need to be a `<d2l-menu-item-link>`? Maybe a regular `<d2l-menu-item>` would work. See [BrightspaceUI/core/components/menu](https://github.com/BrightspaceUI/core/tree/master/components/menu).

media-player.js
```diff
@@ -635,6 +636,20 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		this._sourceType = SOURCE_TYPES.unknown;
 	}
 
+	async _downloadFromSrc() {
+		const response = await fetch(this.downloadSrc);
```
Jevan5, line 640: Other than the status, the response is ignored. Later, in `_onDownloadButtonPress`, the link is fetched (clicking an anchor). Is it possible to add the event handling to `_onDownloadButtonPress`, to listen for an error on the anchor? That way, the request is only made once

qwangd2l: I'll look into this further but from my initial search, there is no handler since it redirects the page/frame to that link when the anchor is clicked.
The idea is to fetch the download source to see if it's valid still before navigating to the download link.

Jevan5: Fetching the URL twice might also be fine, since the browser might cache the result. If it's too difficult to fetch it only once, it shouldn't be too much of an issue to fetch it twice.

rschick: You can change the initial request to a HEAD request. I'm pretty sure content service supports it but if not we can add support for it.

Jevan5 (approved):

qwangd2l:
Since this PR is from my fork, CI won't be able to start Sauce Connect.
I'll open another PR.

PR #76, opened 2021-02-23
fix: check for response status of 206 for downlod-src

qwangd2l:


media-player.js
```diff
@@ -642,7 +642,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 				Range: 'bytes=0-1'
 			}
 		});
-		if (response.status !== 200) {
+		if (response.status !== 200 && response.status !== 206) {
```
rschick, line 645: I think there's a property `response.ok`?

qwangd2l: Yeah, we can change it to use that instead.

rschick (approved):

PR #79, opened 2021-02-25
feat: add new property disable-legacy-download and always emit download-requested event when downloading

qwangd2l:
`disable-legacy-download` to indicate not to use legacy download; instead rely on the event workflow to download.
`download-requested` event is always emitted when downloading.

Related LE PR:
https://github.com/Brightspace/lms/pull/7097

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,7 @@ npm install @brightspace-ui-labs/media-player
 | loop | Boolean | false | If set, once the media has finished playing it will replay from the beginning. |
 | poster | String | null | URL of the image to display in place of the media before it has loaded. |
 | src | String, required |  | URL of the media to play. |
-| download-src | String |  | URL of the media to download. |
-| download-ready | Boolean |  | Set dynamically to perform download using `download-src`. |
+| emit-event-to-download | String |  | If set, the component will emit a `download` event to delegate the download function. |
 
 ```
 <!-- Render a media player with a source file and loop the media when it reaches the end -->
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ npm install @brightspace-ui-labs/media-player
 | loop | Boolean | false | If set, once the media has finished playing it will replay from the beginning. |
 | poster | String | null | URL of the image to display in place of the media before it has loaded. |
 | src | String, required |  | URL of the media to play. |
-| emit-event-to-download | String |  | If set, the component will emit a `download` event to delegate the download function. |
+| emit-event-to-download | String |  | If set, the component will emit a `download-requested` event to delegate the download function. |
 
 ```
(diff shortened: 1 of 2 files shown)


Jevan5 (approved):

Jevan5 (approved):

README.md
```diff
@@ -58,6 +58,7 @@ npm install @brightspace-ui-labs/media-player
 | poster | String | null | URL of the image to display in place of the media before it has loaded. |
 | src | String, required |  | URL of the media to play. |
 | download-src | String |  | URL of the media to download. |
+| download-ready | Boolean |  | Set dynamically to perform download using `download-src`. |
```
Jevan5, line 61: Can this be removed?

qwangd2l: Yes! Thanks for pointing this out.
I've updated the README.md file

README.md
```diff
@@ -57,7 +57,7 @@ npm install @brightspace-ui-labs/media-player
 | loop | Boolean | false | If set, once the media has finished playing it will replay from the beginning. |
 | poster | String | null | URL of the image to display in place of the media before it has loaded. |
 | src | String, required |  | URL of the media to play. |
-| download-src | String |  | URL of the media to download. |
```
rschick, line 60: We still need `download-src` so that file downloads still work in the content tools.

The component should emit an event "download-requested" when `download-src` is empty and `allow-download` is true.

It could actually just always emit the event. If the wrapper isn't listening it will just be ignored.

qwangd2l: I can add back download-src but is it only being used for future work? Because right now the download for content tools go through the `_getDownloadLink` method which adds `attachment=true` to `src`.

rschick: Oh right, sorry I forgot you added it.

README.md
```diff
@@ -57,7 +57,7 @@ npm install @brightspace-ui-labs/media-player
...
+| emit-event-to-download | String |  | If set, the component will emit a `download-requested` event to delegate the download function. |
```
rschick, line 60: Could we use "enable-legacy-download", default true so it continues to work in existing workflows?

rschick: I think it's safe to always emit the download-requested event.. what do you think?

Jevan5: Always emitting the event simplifies the code on both ends. If the event doesn't matter to the outside scope, just ignore it

qwangd2l: Yeah I think it's fine to always emit the `download-requested` event.

I think it would be better if we used a default false property unless we want to add the property in the content tool's render codes or dynamically removing it in LE's MP code.

rschick: Yeah I want to avoid a breaking change that would require updating all the content tools. Maybe `disable-legacy-download` default `false`?

rschick (approved):

PR #85, opened 2021-03-30
Add aria-live attribute to player time div section.

qwangd2l:
Default aria-live for this region to off and when not in focus.
When this region is in focus, set aria-live to polite.


rschick (approved):

rschick:
Could you make sure Rabia is up to speed on how to QA this?

PR #90, opened 2021-07-07
DE44284 - Race condition on src

Jevan5:
The `src` property of the media player could be set after it is dereferenced. In that case, the media player would error out and a loss of functionality would occur.


media-player.js
```diff
@@ -713,7 +713,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 
 	_getDownloadLink() {
 		// Due to Ionic rewriter bug we need to use '_' as a first query string parameter
-		const attachmentUrl = `${this.src}${this.src.indexOf('?') === -1 ? '?_' : ''}`;
+		const attachmentUrl = `${this.src}${this.src?.indexOf('?') === -1 ? '?_' : ''}`;
```
dlockhart, line 716: I _think_ optional chaining should be OK and will get properly transpiled by Babel. It's a really new JS feature though so 🤞 .

dlockhart: This is going to output `'undefined'` though as a string still, is that OK?

Jevan5: Outputting `'undefined'` is okay I think, because this only occurs at the very beginning of the component's lifecycle. This is long before the download button will be clicked and the URL will be used for downloading.

dlockhart (approved):

PR #92, opened 2021-09-23
US131939 - Add quality selector for optional alternative quality sources

muhammadzaidi:
Updated the component to use `source` tags to provide different quality source options e.g.
```
<d2l-labs-media-player>
  <source src="sample-video-144p.mp4" label="SD">
  <source src="sample-video.mp4" label="HD" default>
</d2l-labs-media-player>
```

`src` attribute is used to provide a single quality source, if `src` is present the `source` tags will be ignored and the quality selector is not shown
```
<d2l-labs-media-player src="sample-video.mp4">
```

When there is more than one `source` tag present, a quality selector is shown in the settings menu
Switching to another quality loads the alternative source to the player

Updated the demo to include a compressed 144p version of the sample video to switch to

https://user-images.githubusercontent.com/45248421/134729621-6784ebec-9a68-43ac-841f-5a192acaf704.mp4


demo/index.html
```diff
@@ -17,7 +17,7 @@
 
 			<h2>Video Player</h2>
 			<d2l-demo-snippet>
-				<d2l-labs-media-player src="sample-video.mp4">
+				<d2l-labs-media-player src="sample-video.mp4" altSources="{&quot;SD&quot;: &quot;sample-video-144p.mp4&quot;}">
```
AdmiralCoco, line 20: I would generally avoid having JSON/objects as attributes as it is error prone

Perhaps we should just support SD for now, and HD is the default - i.e., `src=...` is HD, and `sd-src=...` is SD
@MichaelKovalchuk what do you think?

media-player.js
```diff
@@ -872,6 +881,29 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		` : null;
 	}
 
+	_getQualityMenuView() {
+		const theme = this._sourceType === SOURCE_TYPES.video ? 'dark' : undefined;
+		return this.altSources && Object.keys(this.altSources).length > 0 ? html`
+			<d2l-menu-item text="${this.localize('quality')}">
+				<div slot="supporting">${this._selectedQuality || 'HD'}</div>
```
AdmiralCoco, line 888: Can `selectedQuality` be `null` here? It looks like it's defaulted to 'HD' above?

media-player.js
```diff
@@ -872,6 +881,29 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+				<d2l-menu @d2l-menu-item-change=${this._onQualityMenuItemChange} theme="${ifDefined(theme)}">
+					${Object.keys(this.altSources).map(quality => html`
+						<d2l-menu-item-radio
+							?selected="${quality === this._selectedQuality}"
+							text="${`${quality}`}"
```
AdmiralCoco, line 893: Is this quoting required? What happens if we just set `text` to `${quality}`?

demo/index.html
MichaelKovalchuk, line 20: i would think, we do not need both src and altSources at the same time. 
can we make it so if src is a json we parse SD/HD options from there and use the latest as a default?
And if src is not a JSON - we fallback to the existing behaviour

media-player.js
```diff
@@ -1198,6 +1230,34 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 
 	}
 
+	_onQualityMenuItemChange(e) {
+		if (
+			!this.altSources ||
+			!Object.keys(this.altSources) > 0 ||
+			e.target.value === this._selectedQuality ||
+			(e.target.value !== DEFAULT_QUALITY && !(e.target.value in this.altSources))
```
AdmiralCoco, line 1238: Is it possible for `e.target.value` to be something other than a value from altSources (or whatever is in the menu)?

media-player.js
```diff
@@ -1198,6 +1230,34 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+		) return;
+
+		this._selectedQuality = e.target.value;
+
+		const time = this.currentTime;
+		this._media.firstElementChild.setAttribute('src', this._selectedQuality == DEFAULT_QUALITY ? this.src : this.altSources[this._selectedQuality]);
```
AdmiralCoco, line 1244: Missing an extra `=` for strict equality
Also, currently not an issue but if there were other elements added under `this._media`, `firstElementChild` may return something unexpected 

(with the regular `video` tag, there are multiple `source` tags - but our implementation only has one)

media-player.js
muhammadzaidi: It shouldn't be possible but I added the extra check to avoid errors in case the menu's html is manually edited

media-player.js
```diff
@@ -1324,6 +1388,33 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+				if (node.hasAttribute('default')) {
+					this._selectedQuality = quality;
+				} else if (sourceNodes.length === 1) {
+					this._selectedQuality = quality;
+				}
+
+				this._sources[quality] = node.src;
+			}
+		});
+		if (!this._selectedQuality) console.warn("d2l-labs-media-player component requires 'default' attribute on one of the source tags");
```
AdmiralCoco, line 1416: I'm not sure if we want to require this - I'm thinking it just uses the first source (in order) if there are multiple?

demo/index.html
```diff
@@ -17,15 +17,18 @@
 
 			<h2>Video Player</h2>
 			<d2l-demo-snippet>
-				<d2l-labs-media-player src="sample-video.mp4">
+				<d2l-labs-media-player>
```
AdmiralCoco, line 20: Could we add another demo section for the player with single source, through the `src` attribute? And name this one `Video Player (multiple sources)`

media-player.js
```diff
@@ -428,7 +431,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
-		if (!this.src) console.warn('d2l-labs-media-player component requires src text');
+		if (!this.src) {
+			const sourceNodes = Array.from(this.getElementsByTagName('source'));
+			if (sourceNodes.length < 1) console.warn('d2l-labs-media-player component requires source tags');
```
AdmiralCoco, line 436: Nit: I think we should change this to: `d2l-labs-media-player component requires source tags if src is not set`

media-player.js
muhammadzaidi: Changed it to use the first source tag as default if there is no default attribute in any source tags

media-player.js
```diff
@@ -713,7 +724,8 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 
 	_getDownloadLink() {
 		// Due to Ionic rewriter bug we need to use '_' as a first query string parameter
```
MichaelKovalchuk, line 726: i think the comment should be moved one line down here

README.md
```diff
@@ -56,7 +56,7 @@ npm install @brightspace-ui-labs/media-player
 | crossorigin | String | null | If set, will set the `crossorigin` attribute on the underlying media element to the set value.
 | loop | Boolean | false | If set, once the media has finished playing it will replay from the beginning. |
 | poster | String | null | URL of the image to display in place of the media before it has loaded. |
-| src | String, required |  | URL of the media to play. |
+| src | String |  | If set, the single quality source URL of the media to play. `source` tags should be used for multiple quality `sources`. |
```
AdmiralCoco, line 59: Suggestion - update to:
"URL of the media to play. If multiple sources are desired, use `source` tags instead (see below)."

README.md
```diff
@@ -117,6 +117,27 @@ this.document.querySelector('d2l-labs-media-player').addEventListener('loadeddat
 });
 ```
 
+## Multiple qualities using `<source>`
+
+The media player supports switching to different qualities. If any valid `source` tags are present and the `src` attribute on the media player is not present, the `source` tags will be used to provide the different quality sources
```
AdmiralCoco, line 122: Suggestion - update to:
"The media player supports switching to different qualities. If multiple `source` tags are present, a quality selector will be available in the menu. In this case, do not set `src` on `d2l-labs-media-player`."

README.md
```diff
@@ -117,6 +117,27 @@ this.document.querySelector('d2l-labs-media-player').addEventListener('loadeddat
...
+</d2l-labs-media-player
+```
+
+**Attributes**
+
+| Attribute | Type | Default | Description |
+|--|--|--|--|
+| label | String, required | The label for the track, displayed to the user for selection.
+| src | String, required | The URL of the source file.
+| default | Boolean | false | The source to be selected by default. If no source has the `default` attribute then the first `source` tag is selected by default. If multiple `source` tags have the `default` attribute then the latest `<source>` with `default` attribute is used
```
AdmiralCoco, line 137: Suggestion - update to: 
"The source to be selected by default. If no source has the `default` attribute, then the first `source` tag is selected by default. Only one default should be set."

Let's say multiple `default`s are an invalid case, and we don't want to define expected behaviour for it

README.md
```diff
@@ -119,7 +119,7 @@ this.document.querySelector('d2l-labs-media-player').addEventListener('loadeddat
 
 ## Multiple qualities using `<source>`
 
-The media player supports switching to different qualities. If any valid `source` tags are present and the `src` attribute on the media player is not present, the `source` tags will be used to provide the different quality sources
+The media player supports switching to different qualities. If multiple `<source>` tags are present, a quality selector will be available in the menu. In this case, do not set `src` on `d2l-labs-media-player`
```
AdmiralCoco, line 122: Nit: missing period

AdmiralCoco (approved):
Nice work! 👍

PR #95, opened 2021-09-27
Add functionality to search through captions

AdmiralCoco:
- Used [fuse.js](https://fusejs.io/) for searching
- Added webVTT parser to parse webVTT files
- Search only starts with 2 characters
- For this implementation, I did not modify the seek bar, and simply positioned the markers as an overlay
- Tested with a large file (subtitles for a movie) and there was no noticeable performance issue
---
Demo:

https://user-images.githubusercontent.com/6110668/134980253-0c76ec82-dc3b-4bc8-938b-d8f726847fde.mp4


rcausey (approved):

media-player.js
```diff
@@ -53,6 +56,13 @@ const TRACK_KINDS = {
 	subtitles: 'subtitles'
 };
 const Url = URL || window.URL;
+const FUSE_OPTIONS = options => ({
+	keys: ['text'],
+	ignoreLocation: true,
```
MichaelKovalchuk, line 61: why don't we use `findAllMatches` ?

MichaelKovalchuk (approved):

media-player.js
AdmiralCoco: Good question - I didn't try this setting, but based on the description,I figured if the match was found then further searching wouldn't be needed

Is there a case where you think we may need this option?

MichaelKovalchuk: the difference - currently, if we have few perfect matches - it will only show you the first one. I am not entirely assured we need this setting, but considering we want to show many occurrences in the timeline - i propose we use this flag and see how it works

AdmiralCoco: I tested this and thought that it was related to the `includeMatches` parameter
i.e., if the string is `Book 1, Book 2` and `findAllMatches=true`, then the search would return the indexes of both matches

It seems that setting `findAllMatches` didn't affect it though - nor did it affect the current searches

---
>considering we want to show many occurrences in the timeline

To clarify, this is the case - we include all results which contain the searched word, throughout the video

PR #99, opened 2021-09-29
fix: 'trackloaded' event not firing when loading SRT captions files

MarquisLP:
Some applications, such as "Audio/Video Note Captions Editor" in the LMS, depend on the 'trackloaded' event to know when to load in certain UI elements.

Because of that, we need to ensure that 'trackloaded' is fired for both SRT and WebVTT files. This PR adds a missing case for SRT.

This fixes the issue in "Audio/Video Note Captions" where the captions editor shows up empty for Audio/Video Notes that use an uploaded SRT file.

Screen Shot 2021-09-28 at 12 17 34 PM

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -1366,6 +1366,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 				track.cues.forEach(cue => {
 					trackElement.addCue(new VTTCue(cue.start, cue.end, cue.text));
 				});
+				this.dispatchEvent(new CustomEvent('trackloaded'));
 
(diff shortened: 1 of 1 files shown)


MichaelKovalchuk (approved):

MichaelKovalchuk:
do we need to bump the player version?

MarquisLP:
I think it will automatically be bumped due to the commit prefix. From readme:

> Commits which are prefixed with fix: or perf: will trigger a patch release.

AdmiralCoco (approved):

AdmiralCoco:
Actually could you wait to merge this @MarquisLP ? I meant to bump the version in my previous PR

AdmiralCoco:
Sauce tests seem a bit flakey from what I've seen recently - you can try to rerun the job (wouldn't stress it though)

PR #122, opened 2021-11-22
Remove optional chaining because sequences doesn't like it

rcausey:


media-player.js
```diff
@@ -573,7 +573,15 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	get activeCue() {
 		for (let i = 0; i < this._media.textTracks.length; i++) {
 			if (this._media.textTracks[i].mode === 'hidden') {
-				return this._media?.textTracks?.[0].activeCues?.[0] || null;
+				if (this._media &&
```
rcausey, line 576: Can you think of any nicer way to do this?

qwangd2l (approved):

PR #125, opened 2021-11-25
Add thumbnails to the update check and fix thumbnails parsing

clemencetbk:
Previous thumbnails parsing didn't take into account that the URL can contain hyphens e.g. "http://s3.some-bucket/th160h90i5-hash.png"

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -1220,15 +1220,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 			`;
 
 		// format of the thumbnail is [url]/tw<width>h<height>i<interval>-<hash>.[png|jpg]
-		// i.e. we want the first 't' character before the last '-'
-		const thumbArr = this.thumbnails.split('-');
-		const imageName = thumbArr[thumbArr.length - 2].toLowerCase();
+		const imageName = this.thumbnails.split('/').pop();
+		if (!imageName) return;
 
-		const thumbMatch = imageName.match(/t[^t]*$/);
-		if (!thumbMatch) {
-			return;
-		}
-		const thumbStr = thumbMatch[0];
+		const thumbStr = imageName.split('-')[0].toLowerCase();
 		const widthMatch = thumbStr.match(/w(\d+)/);
 		const heightMatch = thumbStr.match(/h(\d+)/);
 		const intervalMatch = thumbStr.match(/i(\d+)/);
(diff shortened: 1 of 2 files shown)


media-player.js
```diff
@@ -1213,8 +1218,17 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
 			`;
-
-		const thumbStr = this.thumbnails.split('-')[0].toLowerCase();
+		
+		// format of the thumbnail is [url]/tw<width>h<height>i<interval>-<hash>.[png|jpg]
+		// i.e. we want the first 't' character before the last '-'
+		const thumbArr = this.thumbnails.split('-');
+		const imageName = thumbArr[thumbArr.length - 2].toLowerCase();
+
+		const thumbMatch = imageName.match(/t[^t]*$/);
```
clemencetbk, line 1227: This parsing will not work if the hash contains hyphens but I don't think it should be an issue given how the thumbnails urls seem to be generated?

muhammadzaidi: Might be simpler to get the filename from the url with `this.thumbnails.split('/').pop()` then `filename.split('-')[0].toLowerCase()` to get the `thumbStr`

MichaelKovalchuk: the only issue with filename.split('-') - we will always rely on the thumbnail file to have a hyphen, which may not be true

clemencetbk: I think it's part of the required format -- don't the thumbnails that get written to s3 follow those requirements?

muhammadzaidi (approved):

media-player.js
```diff
@@ -1214,16 +1219,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
-		const thumbStr = this.thumbnails.split('-')[0].toLowerCase();
-		const widthMatch = thumbStr.match(/w(\d+)/);
-		const heightMatch = thumbStr.match(/h(\d+)/);
-		const intervalMatch = thumbStr.match(/i(\d+)/);
-
-		if (!intervalMatch) return;
-
-		const interval = parseInt(intervalMatch[1]);
-		const thumbWidth = widthMatch ? parseInt(widthMatch[1]) : DEFAULT_PREVIEW_WIDTH;
-		const thumbHeight = heightMatch ? parseInt(heightMatch[1]) : DEFAULT_PREVIEW_HEIGHT;
```
clemencetbk, line 1226: The `th...w...i...(-hash).png` format is guaranteed when the thumbnails name is created so we shouldn't need to fall back on defaults

MichaelKovalchuk (approved):

PR #139, opened 2021-12-06
fix: Recover from URL expiry if paused when URL expires

rcausey:
Fix to issue where URL expiry doesn't resume playback if you're paused when the URL expires and then press play


rcausey:
The issue is that it never gets to a "playing" state if you press play after the URL expires, so it thinks it should be paused after the source is reloaded.

MichaelKovalchuk (approved):

PR #140, opened 2021-12-06
Update getPreference method to return null instead of default value

AdmiralCoco:
- Fixes bug with stringified defaultTrack object

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -898,7 +898,11 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		for (let i = 0; i < this._tracks.length; i++) {
 			const track = this._tracks[i];
 
-			if (track.srclang === this._getSrclangFromTrackIdentifier(this._selectedTrackIdentifier) && track.kind === this._getKindFromTrackIdentifier(this._selectedTrackIdentifier)) return track.label;
+			if (track.srclang === this._getSrclangFromTrackIdentifier(this._selectedTrackIdentifier) &&
+				track.kind === this._getKindFromTrackIdentifier(this._selectedTrackIdentifier)
+			) {
+				return track.label;
+			}
 		}
 
 		return this.localize('off');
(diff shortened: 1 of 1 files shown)


MichaelKovalchuk (approved):

MarquisLP (approved):

media-player.js
```diff
@@ -1616,11 +1616,11 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 			const defaultIgnorePreferences = node.attributes['default-ignore-preferences'];
 			if (node.default || defaultIgnorePreferences) {
 				// Stringified to be parsed in initializeTracks
-				defaultTrack = JSON.stringify({
+				defaultTrack = {
 					srclang: node.srclang,
 					kind: node.kind,
 					ignorePreferences: !!defaultIgnorePreferences,
-				});
+				};
```
MarquisLP, line 1623: For future reference: this is to prevent `defaultTrack.ignorePreferences` from always returning `false` on line 1685. This was causing an issue in Producer where the current language's track would not be enabled, preventing captions from loading.

media-player.js
```diff
@@ -1017,7 +1017,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	}
 
 	_getKindFromTrackIdentifier(trackIdentifier) {
-		return !trackIdentifier ? null : JSON.parse(trackIdentifier).kind;
+		return !trackIdentifier ? null : trackIdentifier.kind;
```
AdmiralCoco, line 1020: Seeing a bug with the code here, going to test again

MarquisLP (approved):

AdmiralCoco:
There are actually a few more places in code where we're treating `this._selectedTrackIdentifier` as a string, which will be affected by this change - going to go over this again.
(e.g., [link](https://github.com/BrightspaceUILabs/media-player/blob/main/media-player.js#L1287))

MichaelKovalchuk (approved):

MarquisLP (approved):

media-player.js
```diff
@@ -1277,14 +1281,19 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+		);
+
 		return this._tracks.length > 0 ? html`
 			<d2l-menu-item text="${this.localize('captions')}">
 				<div slot="supporting">${this._selectedTrackLabel}</div>
 				<d2l-menu @d2l-menu-item-change=${this._onTracksMenuItemChange} theme="${ifDefined(this._getTheme())}">
 					<d2l-menu-item-radio text="${this.localize('off')}" ?selected="${!this._selectedTrackIdentifier}"></d2l-menu-item-radio>
 					${this._tracks.map(track => html`
 						<d2l-menu-item-radio
-							?selected="${this._getTrackIdentifier(track.srclang, track.kind) === this._selectedTrackIdentifier}"
```
AdmiralCoco, line 1287: Removed the usage of `this._getTrackIdentifier` here, as this was creating an object to stringify and compare vs. the (previously stringified) `this._selectedTrackIdentifier`

media-player.js
```diff
@@ -1277,14 +1281,19 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+							?selected="${isTrackSelected(track)}"
 							text="${`${track.label}${track.kind === TRACK_KINDS.captions ? ` (${this.localize('closedCaptionsAcronym')})` : ''}`}"
 							value="${this._getTrackIdentifier(track.srclang, track.kind)}"
```
AdmiralCoco, line 1298: This stringify is needed to get the value set in the HTML

PR #142, opened 2021-12-07
Provide a way to hide captions menu item

rcausey:


rcausey:
I think we can use this in conjunction with `default-ignore-preferences` on the captions track to have the caption enabled by default and not configurable by user

MichaelKovalchuk (approved):

MarquisLP (approved):

AdmiralCoco (approved):

PR #143, opened 2021-12-09
Prevent reloading the same source whenever slots change

rcausey:
I'm not sure if this is causing any issues, but I noticed as I was working on the expiry problem that videos are loaded multiple times, and it's because reloadSource is called for each `<source>` that's changed. This should clean that up. Also, I did tests and couldn't find a reason for that extra code around the initial load that I added before, so I removed it.


AdmiralCoco (approved):

AdmiralCoco:
Not sure about the visual diff - the height of the player (in the screenshot) was changed, is that a result of this PR?

rcausey:
> Not sure about the visual diff - the height of the player (in the screenshot) was changed, is that a result of this PR?

Yeah I can't say I understand why it's changing - but this is going back to the diffs before my original changes: https://github.com/BrightspaceUILabs/media-player/pull/119 - so I'm thinking the different size was due to the multiple loads somehow.

PR #148, opened 2021-12-10
If `src` property is changed, we need to force reload

rcausey:
Previous behaviour only worked if `<source>` slot changed, not if `src` property changed.


media-player.js
```diff
@@ -1815,12 +1815,11 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 		this._sources[quality] = node.src;
 	}
 
-	_reloadSource() {
+	_reloadSource({ forceReload = false } = {}) {
 		if (this._media) {
-
 			const oldSourceNode = this._media.getElementsByTagName('source')[0];
```
rcausey, line 1820: This gets updated automatically when `src` property change, so if condition isn't true

AdmiralCoco (approved):

PR #152, opened 2021-12-15
Workaround for Safari not emitting error on URL expiry

rcausey:


media-player.js
```diff
@@ -1073,7 +1082,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 						@timeupdate=${this._onTimeUpdate}
 						@volumechange=${this._onVolumeChange}
 					>
-						<source src="${this._getCurrentSource()}" @error=${this._onError}>
```
rcausey, line 1076: I needed to standardize how/where the source is getting set, so I'm no longer setting it in the render function. Instead, it always gets set by the reloadSource function.

MichaelKovalchuk (approved):

media-player.js
```diff
@@ -1812,15 +1821,30 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+				// around this, we manually fire an error slightly before the URL is expected
+				// to expire. Note that this only currently works with URLs that include the
+				// "Expires" query parameter (e.g. CloudFront signed URLs). It's also not a
+				// good long-term solution since it depends on a somewhat accurate client
+				// system time.
+				if (isSafari()) {
+					const expires = parseUrlExpiry(updatedSource);
+					if (expires) {
+						const timeToExpiry = (expires * 1000) - Date.now() - SAFARI_EXPIRY_EARLY_SWAP_SECONDS;
+						const timeoutPeriod = Math.max(timeToExpiry, SAFARI_EXPIRY_MIN_ERROR_EMIT_SECONDS);
```
rcausey, line 1843: Trying to prevent a case where the client system time is off and we just keep reloading over and over. With SAFARI_EXPIRY_MIN_ERROR_EMIT_SECONDS, we at least wait that amount of time before attempting a reload.

media-player.js
```diff
@@ -65,6 +65,14 @@ const DEFAULT_PREVIEW_WIDTH = 160;
 const DEFAULT_PREVIEW_HEIGHT = 90;
 const DEFAULT_LOCALE = 'en';
 
+const SAFARI_EXPIRY_EARLY_SWAP_SECONDS = 10;
```
rcausey, line 68: Swap 10 seconds before expiry (accounting for very slight differences in system time)

PR #156, opened 2022-01-06
Update poster implementation

muhammadzaidi:
Updated the `poster` functionality to use a custom implementation instead of using the `<video>` `poster` attribute, fixes the inconsistencies across browsers and mobile from using the `<video>` `poster` attribute


media-player.js
```diff
@@ -155,11 +156,34 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
...
+			}
+
+			#d2l-labs-media-player-video-poster-play-button {
+				background-color: rgba(0, 0, 0, 0.69);
+				border: none;
+				border-radius: 50%;
+				cursor: pointer;
+				padding: 2em;
+				position: absolute;
+				z-index: 3;
```
AdmiralCoco, line 174: Nit: I think this can be the same z-index as the controls? (for consistency)

AdmiralCoco (approved):

rcausey:
This looks good. I'm just debating whether we need that extra play button. Is it necessary for any reason?

AdmiralCoco (approved):

muhammadzaidi:
> This looks good. I'm just debating whether we need that extra play button. Is it necessary for some reason?

@rcausey I added the play button because in iOS it doesn't show the controls so it was just showing the poster image

rcausey:
Got it, makes sense. Did you see any issues in iOS or Safari on desktop? I think the issues we had before was that you needed to click play twice to get it to play and it would show the pause button even though it didn't start playing the video.

muhammadzaidi:
> Got it, makes sense. Did you see any issues in iOS or Safari on desktop? I think the issues we had before was that you needed to click play twice to get it to play and it would show the pause button even though it didn't start playing the video.

Yeah those issues seem fixed now

rcausey (approved):

PR #159, opened 2022-01-18
Lit 2 Upgrade: backwards-compatible changes

dlockhart:
This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
- [x] audit `ifDefined` usages for null parameters

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.

### DIFF
--- a/media-player-audio-bars.js
+++ b/media-player-audio-bars.js
@@ -1,6 +1,6 @@
-import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { css, html, LitElement } from 'lit-element';
 import ResizeObserver from 'resize-observer-polyfill';
-import { styleMap } from 'lit-html/directives/style-map';
+import { styleMap } from 'lit-html/directives/style-map.js';
 
 const AUDIO_BARS_GRADIENTS = [
 	{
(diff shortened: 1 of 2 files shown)


media-player-audio-bars.js
```diff
@@ -1,6 +1,6 @@
-import { css, html, LitElement } from 'lit-element/lit-element.js';
+import { css, html, LitElement } from 'lit-element';
```
dlockhart, line 1: This change wasn't actually necessarily, but I did it for consistency with other repos.

media-player.js
```diff
@@ -918,6 +923,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	}
 
 	get _media() {
+		if (!this.shadowRoot) return null;
```
dlockhart, line 926: This was the source of most of the other changes. Because `this.shadowRoot` isn't available until after the component has been connected, it's possible that this getter is called before that. So I've put in guards around all the places that access `this._media` (some of which existed already).

I didn't put guards around places that seemed like they'd be safely run after connection -- like event listeners on the underlying DOM nodes.

annabelsegalld2l (approved):

PR #161, opened 2022-02-09
Disable setting preferences in Producer X

lchan131:
Changing preferences (volume, playback speed, etc) in Advanced Editing should not not set/save preferences.


rcausey (approved):
See my other comment about versioning in the PR

PR #167, opened 2022-03-28
adding support for Hindi

dlockhart:
While Hindi got added by Serge, because this is manually listing out each language it also needs to be added here.

I'd highly recommend switching this repo over to the `LocalizeDyanmicMixin` from core, as it will work with any future language without any changes.

Separately I'm going to update the CODEOWNERS file to a team, since having individuals in it isn't great.


rcausey (approved):
Thanks Dave, I'll get someone to work on switching to `LocalizeDynamicMixin` soon.

PR #169, opened 2022-04-06
Change to LocalizeDynamicMixin

lchan131:


media-player.js
```diff
@@ -20,8 +20,8 @@ import { FocusVisiblePolyfillMixin } from '@brightspace-ui/core/mixins/focus-vis
 import fullscreenApi from './src/fullscreen-api.js';
 import Fuse from 'fuse.js';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
-import { InternalLocalizeMixin } from './src/mixins/internal-localize-mixin.js';
```
rcausey, line 23: do we still need this mixin?

media-player.js
```diff
@@ -638,6 +638,14 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalLocalizeMixin(RtlMix
 	get isIOSVideo() {
 		return IS_IOS && this.mediaType === SOURCE_TYPES.video;
 	}
+	static get localizeConfig() {
```
qwangd2l, line 641: I'd prefer if we moved the static methods together if we're defining this within the component file.
The other option would be to refactor this into a mixin and use it like the old `InternalLocalizeMixin`.

media-player.js
lchan131: No. It has been removed.

media-player.js
rcausey: Yeah good point. I think keeping it as a mixin is a good idea rather than implementing directly in this file.

rcausey: See: https://github.com/Brightspace/content-components/blob/5ad01f515cd6dd5446b868739a0f92f8e03de535/core/d2l-content-renderer/src/mixins/internal-localize-mixin.js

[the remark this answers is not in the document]
lchan131: Lint rule added

lchan131: Moved to its own class

qwangd2l:
Looks good to me, but you might want to consider removing the `internal-localize-mixin.js` file since it's no longer used (it's what Rhys commented about earlier).

media-player.js
lchan131: That is strange. I am seeing it is highlighted red, so I'm positive it is removed. Try refreshing.

qwangd2l: The file, or the import statement? I think we're talking about the removal of the file.

rcausey (approved):
Looks good. Make sure to bump the version on merge.

qwangd2l (approved):

PR #180, opened 2022-08-05
Handle error on url parse

MichaelKovalchuk:


media-player.js
```diff
@@ -2017,7 +2021,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 				// good long-term solution since it depends on a somewhat accurate client
 				// system time.
 				if (isSafari() && updatedSource !== undefined) {
-					const expires = parseUrlExpiry(updatedSource);
+					const expires = tryParseUrlExpiry(updatedSource);
```
MichaelKovalchuk, line 2024: `expires` only used for videos added to Content Service.
This error is thrown by the videos added to Classic Content in Safari

MarquisLP (approved):

dlockhart (approved):
Thanks for the quick fix!

media-player.js
MichaelKovalchuk:
```
sharingOrgUnit
legacy
topic
placement
```
are the values I found so far

PR #181, opened 2022-08-10
Fix slider not zooming on drag

lchan131:


media-player.js
```diff
@@ -701,6 +701,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 		return this._media ? this._media.textTracks : null;
 	}
 
+	get zoomBar() {
```
MichaelKovalchuk, line 704: i would make it a function that returns zoomBar

media-player.js
```diff
@@ -2069,8 +2072,8 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 	}
 
 	_sliderChange() {
-		if (!this._zoomBar) return;
-		const zoomBarValue = this._zoomBar.immediateValue - 50;
+		if (!this.zoomBar) return;
```
MichaelKovalchuk, line 2075: if function is used - you can do something like this: 
const zoomBar = this.getZoomBar() or, alternatively:
```
const zoomBar = this.shadowRoot.getElementById('d2l-labs-media-player-zoom-bar');
if (!zoomBar) return;
const zoomBarValue = zoomBar.immediateValue - 50;
```

media-player.js
lchan131: I just removed this getter

MichaelKovalchuk (approved):

media-player.js
lchan131: I just took your advice and simply called getElementById.

PR #177, opened 2022-07-25
Support mixed video

lchan131:
* Support mixed video in metadata.
* Add zooming for mixed video
### With layout preset
https://user-images.githubusercontent.com/98043376/181117404-ed203330-99d4-49ff-b5f0-1794087f9fd9.mov

### Without layout preset
https://user-images.githubusercontent.com/98043376/181117355-3958f3dc-7051-4faa-9f76-50556ef3c65a.mov

PR #185, opened 2022-09-13
US142768: Bring back Mixed Layouts slider with fixes for ISF cutoff and zooming

MarquisLP:
This PR reintroduces the Mixed Layouts slider (#177) alongside some fixes.

## ISF Cutoff Fix

In ISF and some other contexts, the changes from #177 caused the video to grow out of bounds, cutting off the visible area and pushing the controls outside the Media Player's container, making them inaccessible:

Media Player Mixed Layouts ISF Fix - Before

By adding `max-height` back to the CSS for `#d2l-labs-media-player-media-container`, the cut off is fixed and the Media Player in ISF now matches what we currently have on Staging. (Note that the slight cut off on the bottom is also present on Staging.)

Media Player Mixed Layouts ISF Fix - After

## Zooming and Positioning Fix

The previous implementation of the slider's "zoom" logic had a bug: when the scaling crosses a certain threshold, the video stops being enlarged and instead it is pushed off to the side. This causes the Media Player to appear as if it is zooming in on the bottom-left corner of the video where there is empty black space.

Below is a video showing this issue. Note that I temporarily removed `overflow: hidden` on the video container and pushed the `<video>` and slider to higher z-indexes, in order to make it clearer to see what's happening.

https://user-images.githubusercontent.com/9592685/189959607-17af19d1-493b-43b8-93c3-bee3fb2df13f.mp4

This issue was happening in all browsers, and appears to be caused by a limitation of scaling videos with `width` and `height`.

To fix this issue, I reimplemented the slider's logic using the CSS `transform(scale(...))` function. There is no limitation on scaling using this function, and it allows us to simplify the logic a bit. Below are demo videos showing the new slider implementation.

**With a Layout Preset**

https://user-images.githubusercontent.com/9592685/189959786-1dfff71c-d0e4-4df6-90e9-cf1ee3d6f18a.mp4

**With No Layout Preset**

https://user-images.githubusercontent.com/9592685/189959806-60f26978-f182-4a49-ac3f-ce5606a6687c.mp4


media-player.js
```diff
@@ -953,6 +1028,74 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
...
+		const zoomedFrameHeight = zoomedFrame.bottom - zoomedFrame.top + 1;
+		const otherFrameWidth = otherFrame.right - otherFrame.left + 1;
+		const otherFrameHeight = otherFrame.bottom - otherFrame.top + 1;
+
+		this.zoomLevel = zoomLevel;
+		const zoom = otherFrameWidth / zoomedFrameWidth * Math.abs(this.zoomLevel) / SLIDER_STEPS;
+		const zoomPercentage = zoom * 100;
+
+		// The video will be enlarged using the CSS transform(scale(...)) property.
+		// When scaling, the origin point will be the center point of the video.
```
MarquisLP, line 1052: I had tried using `transform-origin` set to the center point of the zoomed frame, thinking that this would cause the video to automatically keep that center point in focus while scaling. However, this was not the case, and attempting to fix the positioning proved to be quite complicated.

Because of that, I decided to keep the `transform-origin` at the default of `center` in order to keep the calculations straightforward.

MichaelKovalchuk (approved):

media-player.js
```diff
@@ -676,23 +728,6 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
...
-					this._heightPixels = Math.floor(Math.max(height, this._lastHeight));
-				} else {
-					this._heightPixels = null;
-				}
-
-				this._twoHeightsAgo = this._lastHeight;
-				this._lastHeight = height;
-
-				this._twoWidthsAgo = this._lastWidth;
-				this._lastWidth = width;
```
MarquisLP, line 695: Removed this because setting an explicit height on `#d2l-labs-media-player-zoom-bar-container` conflicts with the new `aspect-ratio` attribute. (If `height` is explicitly set, then `aspect-ratio` is ignored.)

In theory, the flickering should no longer occur because `<video>` is now absolutely positioned, so it shouldn't be affected by temporary fluctuations in the positioning of other elements. I have tested these changes on all browsers and haven't encountered any flickering. However, we can bring back (and possibly rework) this logic later if we do find issues.

PR #186, opened 2022-09-14
US142768: Fix issues with empty Mixed Layout frames, flickering, and Mixed Layout zoom calculations

MarquisLP:
This PR fixes several issues discovered since #185.

## Mixed Layout Slider unresponsive when layout frames are empty

It's possible for the frames array in layout presets to be empty. Example:

```json
"layoutPresets": {
  "frames": []
}
```

In this case, the slider should default to the basic zoom functionality. Before this fix, the slider would do nothing at all in this scenario.

## Flickering on Windows

Before #185, we had code that prevented the Media Player from rapidly resizing a few pixels on Windows, producing a flickering effect. It was removed in #185 due to conflicts with other code, but it turns out that this was untrue (see comments below). Additionally, we determined that this fix is still needed because the flickering was happening in Media Library.

## Incorrect zooming for Mixed Layout Slider on Capture App videos

The calculations for the Mixed Layouts Slider are off -- specifically, when zooming in all the way on one side, the Y-axis translation is inaccurate and cuts off portions of the video height-wise. This is most apparent in videos from Capture App, where half the video is cut off when zoomed in all the way.


MarquisLP:
Added back the flickering fix for Windows that was removed in #185. Contrary to what was said in that PR, The "height" value that we set on `<video>` does not conflict with "aspect-ratio", because "aspect-ratio" is set on the container div rather than the `<video>` element. I've confirmed that this fixes the flickering issue on Windows and does not break the aspect ratio sizing.

Kevin identified a separate issue involving the Mixed Layouts Slider not zooming into the right area for videos published from Capture App. I will rework the zoom calculations and include them in this PR as well.

MichaelKovalchuk (approved):

MarquisLP:
Another issue: in Producer, wide videos cause the right-hand controls to become cut off:

Screen Shot 2022-09-14 at 6 06 03 PM

MarquisLP:
Merging this PR now so that we can get the flickering fix on Staging. The Producer issue noted above will be addressed in a followup PR.

PR #199, opened 2022-10-11
Fix scrolling issue

clemkoh:
With `scrollIntoView`, the window would scroll back to the video when it was still playing even if the user had scrolled away from it. 

scroll

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -633,7 +633,6 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 			#video-transcript-viewer {
 				bottom: 55px;
 				color: white;
-				overflow-anchor: none;
 				overflow-y: auto;
 				position: absolute;
 				right: 0;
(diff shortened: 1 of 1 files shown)


media-player.js
```diff
@@ -1673,8 +1673,22 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
...
+				this._transcriptViewer = this.shadowRoot.getElementById('video-transcript-viewer')
+					|| this.shadowRoot.getElementById('audio-transcript-viewer');
+			}
 			const cue = this.shadowRoot.getElementById('transcript-viewer-active-cue');
-			cue?.scrollIntoView({ behavior: 'smooth', block: 'center', inline: 'nearest' });
+			const cueRect = cue.getBoundingClientRect();
+			const transcriptRect = this._transcriptViewer.getBoundingClientRect();
+			if (cue && this.activeCue && this._canScroll) {
+				if (cueRect.bottom + 50 > transcriptRect.bottom && cueRect.height <= transcriptRect.height) {
+					this._scrolling = true;
```
MarquisLP, line 1684: Doesn't seem like `this._scrolling` is used anywhere -- can it be removed?

MarquisLP (approved):

PR #200, opened 2022-10-12
Move update cue logic to _onCueChange()

clemkoh:
Fixes an issue where the position of the active cue would get updated before the text

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -722,6 +722,8 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 		};
 		this._videoStyle = {};
 		this._zoomLevel = 0;
+		this.afterCaptions = [];
+		this.beforeCaptions = [];
 	}
 
 	get currentTime() {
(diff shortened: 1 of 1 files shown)


media-player.js
```diff
@@ -1672,14 +1674,52 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 
 	_onCueChange() {
 		if (this.transcriptViewerOn) {
+			let cues = null;
+			for (let i = 0; i < this._media.textTracks.length; i += 1) {
+				cues = this._media.textTracks[i]?.cues;
+				if (cues) {
+					const activeCues = this._media.textTracks[i].activeCues;
+					if (!activeCues) break;
+					this.transcriptCue = activeCues[activeCues.length - 1];
```
MarquisLP, line 1683: Nit: we could rename `transcriptCue` to `transcriptActiveCue` so that its purpose is clearer in other parts of the code.

media-player.js
```diff
@@ -1672,14 +1674,52 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 
 	_onCueChange() {
 		if (this.transcriptViewerOn) {
```
MarquisLP, line 1676: Nit: now that `_onCueChange()` handles two things (updating the transcript viewer and displaying the active captions cue), could we separate them into helper methods? Something like `_updateTranscriptViewerCues()` might be good for the former.

media-player.js
```diff
@@ -2219,51 +2258,19 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
...
-			}
-			if (!defaultTrack) defaultTrack = this._media.textTracks[0];
-			defaultTrack.mode = 'hidden';
-			this._selectedTrackIdentifier = { kind: defaultTrack.kind, srclang: defaultTrack.language };
-			this.requestUpdate();
-			return;
-		}
 
-		this.locale = transcriptLocale?.toLowerCase();
+		this._setZoomLevel(0);
```
MarquisLP, line 2262: Out of curiosity, why is the slider being reset to 0 here? It might be good to add a comment explaining this.

MarquisLP (approved):

media-player.js
```diff
@@ -633,6 +633,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 			#video-transcript-viewer {
 				bottom: 55px;
 				color: white;
+				overflow-anchor: none;
```
clemkoh, line 636: This is needed as there is a feature in Chrome where the elements will auto-scroll inside transcript viewer. So without this attribute set to none, it scrolls a little bit at every caption instead of scrolling "in batches".

media-player.js
```diff
@@ -2113,6 +2113,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 				this._media.textTracks[i].mode = 'disabled';
 			}
 		}
+		this._updateTranscriptViewerCues();
```
clemkoh, line 2116: Update when the language is changed, otherwise the transcript cues are not updated if there is no active cue

media-player.js
```diff
@@ -2422,13 +2422,16 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
...
-			cues = this._media.textTracks[i]?.cues;
-			if (cues) {
-				const activeCues = this._media.textTracks[i].activeCues;
-				if (!activeCues) break;
-				this.transcriptActiveCue = activeCues[activeCues.length - 1];
-				break;
+			const currTracks = this._media.textTracks[i];
+			if (currTracks?.cues) {
+				const activeCues = currTracks.activeCues;
+				if (lang === currTracks.language) {
```
clemkoh, line 2431: Use selected language instead of active cues to find which set of cues we should display. Not sure if this would actually cause any issues but it seems better to have a single source of truth

clemkoh: Sorry for all the commits haha this should be the last thing

MarquisLP: This should be fine -- I think the previous implementation would be better if we had to display multiple active tracks at once, but the Media Player only ever displays one track at a time for the selected language. So using `this._selectedTrackIdentifier` is alright.

MarquisLP (approved):

PR #201, opened 2022-10-18
Add attribute for selected language

clemkoh:
Media player will use selectedTrackSrcLang to get the download url for transcript/captions: https://github.com/Brightspace/content-components/pull/564

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -101,6 +101,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 			mediaType: { type: String, attribute: 'media-type' },
 			metadata: { type: Object },
 			poster: { type: String },
+			selectedTrackSrcLang: { type: String },
 			src: { type: String },
 			thumbnails: { type: String },
 			disableSetPreferences: { type: Boolean, attribute: 'disable-set-preferences' },
@@ -1060,6 +1061,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 		if (changedProperties.has('thumbnails')) {
 			this._getThumbnails();
 		}
+
+		if (changedProperties.has('_selectedTrackIdentifier')) {
+			this._getSelectedTrackSrclang();
+		}
 	}
 
 	exitFullscreen() {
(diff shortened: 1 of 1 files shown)


media-player.js
```diff
@@ -101,6 +101,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 			mediaType: { type: String, attribute: 'media-type' },
 			metadata: { type: Object },
 			poster: { type: String },
+			selectedTrackSrcLang: { type: String },
```
MarquisLP, line 104: Is this attribute going to be passed in from the Media Player's parent components? Or is its value only set by the Media Player itself?

If it's the latter, then I think that instead of making it a public attribute, we could make it a private attribute and expose it via a `get` function like [the one we have for `currentTime`](https://github.com/BrightspaceUILabs/media-player/blob/a5dc61bc3f62ec94930f4ca86ad337515ab5964c/media-player.js#L732).

media-player.js
```diff
@@ -785,6 +786,10 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 		return (this._media && this._media.paused);
 	}
 
+	get selectedTrackSrcLang() {
+		return this._selectedTrackSrcLang;
+	}
+
```
MarquisLP, line 792: Thinking on this some more, I think we can have this getter return `this._getSrclangFromTrackIdentifier(this._selectedTrackIdentifier)` directly and remove `_selectedTrackSrcLang`. Since `_selectedTrackIdentifier` is an attribute, the getter's value should automatically update whenever that value changes.

media-player.js
```diff
@@ -101,6 +101,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 			mediaType: { type: String, attribute: 'media-type' },
 			metadata: { type: Object },
 			poster: { type: String },
+			_selectedTrackSrcLang: { type: String, attribute: false },
```
MarquisLP, line 104: I think we can remove this as it's unused now.

MarquisLP (approved):

PR #206, opened 2022-12-05
Remove npm token in order to publish to public npm

will-malisch:


gordon-plumb (approved):

MichaelKovalchuk:
why we are publishing to public npm?

PR #208, opened 2022-12-05
Undo changes related to code artifact and match-lms-release

will-malisch:
From what I understand about code artifact packages, they need to be prefixed with @d2l/ so I went with what other private BrightspaceUILabs repos were doing and prefixed the title with @d2l/labs-media-player. Let me know if this works.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,11 @@
+
 name: Release
 on:
   push:
-    branches: [main, 'release/**']
+    branches:
+      - main
+      - '[0-9]+.x'
+      - '[0-9]+.[0-9]+.x'
 jobs:
   release:
     name: Release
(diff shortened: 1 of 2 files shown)


will-malisch:
UPDATE: looks like there actually are a lot of other repos that I just doing @d2l/<title-of-repo> instead of @d2l/labs-<title-of-repo>. Going to go with the first approach because this will be easier for people trying to install this going forward, and for updating all the references

gordon-plumb (approved):

MichaelKovalchuk (changes requested):
I am not sure we can not merge this PR because the repo must be private first, otherwise the CODEARTIFACT_AUTH_TOKEN, AWS Secrets are not accessible.
I would suggest chatting with Web Platform to figure out the best steps. Possibly:
* Move the code to the `Brightspace` organization media-player private repo
* Use code artifact auth token for publishing the packages

will-malisch:
https://d2l.slack.com/archives/C0PHG3QB0/p1670268048923839

Note: we can put this in BrightspaceUI, it also has AWS/Code artifact secrets

rcausey:
Chatting with @MichaelKovalchuk, let's hold on this

PR #212, opened 2023-01-20
add back isf cut off fix

gordon-plumb:
Removed this fix with the layout slider removal but it is still needed because aspect-ratio is still being set on the media container div. Adding it back prevents videos from being cut off in ISF when the aspect ratio doesn't match the dimensions of the iframe. Need to keep aspect-ratio because transcript viewer relies on it to maintain the size of the player when switching between views.


gordon-plumb:
I think this will require a late add.

MarquisLP (approved):

PR #213, opened 2023-01-20
Adding duration-hint

ktonon:
Measured in seconds, this will be used if the duration determined by the video element's durationchange event is either infinity or not a number. This can happen when using a url for an video which has just been recorded but not uploaded (i.e. a blob)

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -85,6 +85,7 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 			allowDownload: { type: Boolean, attribute: 'allow-download', reflect: true },
 			autoplay: { type: Boolean },
 			crossorigin: { type: String },
+			durationHint: { type: Number, attribute: 'duration-hint' },
 			hideCaptionsSelection: { type: Boolean, attribute: 'hide-captions-selection' },
 			hideSeekBar: { type: Boolean, attribute: 'hide-seek-bar' },
 			locale: { type: String },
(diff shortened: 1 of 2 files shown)


media-player.js
```diff
@@ -1582,6 +1583,9 @@ class MediaPlayer extends FocusVisiblePolyfillMixin(InternalDynamicLocalizeMixin
 
 	_onDurationChange(e) {
 		this._duration = e.target.duration;
+		if ((!isFinite(this._duration) || isNaN(this._duration)) && this.durationHint) {
+			this._duration = this.durationHint;
+		}
```
MarquisLP, line 1588: `this._duration = e.target.duration` might trigger an update of the component before it gets to the code below. To prevent update inconsistencies, we might want to set `this._duration` only once in this function:

```javascript
const newDuration = e.target.duration;
if ((!isFinite(newDuration)) || ...) {
  this_duration = this.durationHint;
} else {
  this._duration = newDuration;
}
```

ktonon:
Sorry, I'll revert the whitespace changes...

MarquisLP (approved):

ktonon:
Should this be a feature or a bug fix? I'm thinking just bug fix even though it adds a property to the component

MarquisLP:
> Should this be a feature or a bug fix? I'm thinking just bug fix even though it adds a property to the component

Sorry for late reply, but I think `feature` is appropriate since the new property is publicly-accessible outside of the component.

PR #214, opened 2023-03-02
feat: Use getFocusPseudoClass for focus-visible

eKoopmans:
👋 Hi! I'm from team Gaudi.

We [recently added a new approach](https://github.com/BrightspaceUI/core/pull/3215) for the `:focus-visible` selector, and are in the process of phasing out `FocusVisiblePolyfillMixin`. This PR replaces its usage with the new `getFocusPseudoClass`, which returns `:focus-visible` on browsers that support it, and `:focus` on those that don't.

The main benefits here are switching to browser-native tech and removing the overhead of the polyfill, while still having a fallback for the [browsers that don't support focus-visible](https://caniuse.com/?search=focus-visible).

Let me know if you have any questions, thanks!

Rally: https://rally1.rallydev.com/#/?detail=/userstory/685236909367&fdp=true


eKoopmans:
Visual-diff changes look okay to me, there's a large difference on the `video` tests but I think it's just some video rendering difference, or a slightly different frame? Merging them.

dbatiste (approved):

ktonon (approved):

PR #240, opened 2023-07-21
Single lang chapters

MichaelKovalchuk:
Locale was never passed into the player
Because of that, we only ever supported english chapters
Going forward, we will be using a single lang chapters but the existing english localized chapters will continue to work until the migration is complete

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,6 @@ npm install @brightspace-ui-labs/media-player
 | duration-hint | Number | 1 | Measured in seconds. If set and the duration cannot be determined automatically, this value will be used instead.
 | hide-captions-selection | Boolean | false | If set, the menu item to configure captions is hidden. |
 | hide-seek-bar | Boolean | false | If set, the seek bar will not be shown. |
-| locale | String | en | If set, will display chapter titles in the given locale when possible. |
 | loop | Boolean | false | If set, once the media has finished playing it will replay from the beginning. |
 | media-type | ["video", "audio"] | null | Whether the video or audio player should be rendered. If not set, a loading indicator will be displayed until set.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -51,7 +51,6 @@ npm install @brightspace-ui-labs/media-player
 | duration-hint | Number | 1 | Measured in seconds. If set and the duration cannot be determined automatically, this value will be used instead.
 | hide-captions-selection | Boolean | false | If set, the menu item to configure captions is hidden. |
 | hide-seek-bar | Boolean | false | If set, the seek bar will not be shown. |
-| locale | String | en | If set, will display chapter titles in the given locale when possible. |
```
MichaelKovalchuk, line 54: Locale was never passed in. 
Removing this parameter as unnecessary for the single lang chapters

gordon-plumb (approved):

media-player.js
```diff
@@ -1157,17 +1150,15 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
...
-			return chapterTitle[this.locale];
+		if (typeof chapterTitle === 'string') {
+			return chapterTitle;
 		}
+
 		for (const locale in chapterTitle) {
-			if (this.locale.split('-')[0] === locale.split('-')[0]) {
+			if (locale.split('-')[0] === 'en') {
 				return chapterTitle[locale];
 			}
```
MarquisLP, line 1160: Just to clarify: this is for backwards compatibility with older videos that still have multi-lang chapters?

MarquisLP (approved):

media-player.js
MichaelKovalchuk: yes. we only ever supported english chapters. so this is to insure backwards compatibility

PR #243, opened 2023-11-21
Make Media Player accessibility compliant

LukeDasios:
Context: folio-app consumes the media-player component and uses it within one of our own components (video-renderer). When creating aXe tests for our video-renderer component I encountered this error:

image

And I believe that my changes will remedy this

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -1333,7 +1333,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 		if (!this.poster || this.autoplay || !this._posterVisible) return;
 
 		const playIcon = !this._loading ? html`
-			<button id="d2l-labs-media-player-video-poster-play-button" title=${this.localize('playButton')} aria-label=${this.localize('playButton')} transcript="${ifDefined(this.transcriptViewerOn ? true : undefined)}"
+			<button id="d2l-labs-media-player-video-poster-play-button" title=${this.localize('play')} aria-label=${this.localize('play')} transcript="${ifDefined(this.transcriptViewerOn ? true : undefined)}"
 				@click=${this._onVideoClick}>
 				<d2l-icon icon="tier1:play" theme="${ifDefined(this._getTheme())}"></d2l-icon>
 			</button>
--- a/media-player.js
+++ b/media-player.js
@@ -1333,7 +1333,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 		if (!this.poster || this.autoplay || !this._posterVisible) return;
 
 		const playIcon = !this._loading ? html`
-			<button id="d2l-labs-media-player-video-poster-play-button" title="Button" transcript="${ifDefined(this.transcriptViewerOn ? true : undefined)}"
+			<button id="d2l-labs-media-player-video-poster-play-button" title=${this.localize('playButton')} aria-label=${this.localize('playButton')} transcript="${ifDefined(this.transcriptViewerOn ? true : undefined)}"
 				@click=${this._onVideoClick}>
 				<d2l-icon icon="tier1:play" theme="${ifDefined(this._getTheme())}"></d2l-icon>
 			</button>
--- a/media-player.js
+++ b/media-player.js
@@ -1333,7 +1333,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 		if (!this.poster || this.autoplay || !this._posterVisible) return;
 
 		const playIcon = !this._loading ? html`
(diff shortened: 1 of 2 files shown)


media-player.js
```diff
@@ -1333,7 +1333,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 		if (!this.poster || this.autoplay || !this._posterVisible) return;
 
 		const playIcon = !this._loading ? html`
-			<button id="d2l-labs-media-player-video-poster-play-button" transcript="${ifDefined(this.transcriptViewerOn ? true : undefined)}"
+			<button id="d2l-labs-media-player-video-poster-play-button" title="Button" transcript="${ifDefined(this.transcriptViewerOn ? true : undefined)}"
```
rcausey, line 1336: Should it be more descriptive? Like "Play button"? And should it be localized?

eKoopmans: Yes to both, and also we should be using an aria label, not `title`.

eKoopmans: Left some relevant comments on this other PR:

https://github.com/Brightspace/d2l-pdf-viewer/pull/48/files#r1401075055

LukeDasios: For whatever reason I need to use title to get the aXe tests to pass, but I can include an aria-label too. I checked and this repo is using TIE, so my lang term changes will be made in a week or so and when it does I'll get assigned to the PR and will merge it

dbatiste: Note: the role ("button") is part of the a11y object model and automatically announced. So, the label should be "Play" or something more descriptive, not including "Button".

Our core button components either place the text within the `<button>Play</button>` or include the `title` (in the case of `d2l-button-icon`) as well, which is probably why our aXe tests pass. When there is no visible text (like `d2l-button-icon`), the native browser tooltip that is displayed using the `title` can be very helpful for users who do not recognize the icon. I've seen popular websites omit this, effectively making their icon buttons "mystery buttons", which is terrible.

rcausey: I wonder if it makes sense to use the existing "Play" langterm?

dbatiste: Probably yeah. If you mean from the LMS, I've definitely done that - when I do, I usually copy the existing translations. I'm not sure if it saves us any translation cost though, or if it just sets up the translations from the get-go rather than waiting for the whole translation process fill them in.

rcausey: Thanks Dave. We have a "Play" langterm already defined in this repo: https://github.com/BrightspaceUILabs/media-player/blob/27e123ec3bf93b6d05340f49652a766f377316fe/lang/en.js#L18

LukeDasios: 👀 Thanks for pointing that out @rcausey. See the commit to change the langterm to use it here: 9ecc069b27782d3f35affff9b10b004c626258dd

eKoopmans: To Dave's comment, I see that `d2l-button-icon` uses both `aria-label` and `title`:
https://github.com/BrightspaceUI/core/blob/main/components/button/button-icon.js

@dbatiste When would you recommend using both, vs just `aria-label`? I'm curious which you'd recommend for this progress-bar PR:
https://github.com/Brightspace/d2l-pdf-viewer/pull/48

eKoopmans (approved):

MichaelKovalchuk (approved):

media-player.js
dbatiste: Good question. If we assume screenreader users are already covered by the `aria-label`, then I'd say any time it might not be obvious to sighted/non-screenreader users what an element doses. Actionable elements that don't have text come to mind. I don't necessarily think we're consistent on this. 

The `d2l-tabs` uses `title` [here](https://github.com/BrightspaceUI/core/blob/main/components/tabs/tabs.js#L344) even though the icons are probably pretty obvious. On the other hand, `d2l-switch` does not [here](https://github.com/BrightspaceUI/core/blob/main/components/switch/switch-mixin.js#L202), even when it's label text is hidden, and it's visible label is problematic for other reasons, but I could make a pretty strong argument for giving it a `title` when the label is hidden. Having said that, the progress element [here](https://github.com/BrightspaceUI/lms-core/blob/main/src/components/files/file-input-progress.js#L96) does not use a `title` - it is obvious that it's a progress bar, and the visual context should be enough for a user to know what it's for, but one could make an argument for giving it a `title`.

A slight aside: I notice that the progress bar you linked to is not using the native `progress` element. I wonder if it could be. I'm not sure the history there, but it's an older component using Polymer and fastdom. I imagine it could be simplified a bunch. It seems we have a bunch of different sets progress styles, and it would also be good to get some alignment on them.

eKoopmans: Thanks Dave! Yep I agree that the linked progress bar could be greatly overhauled, not something we have capacity for right now though. From your explanation, I think we're safe to omit `title` on the progress bar.

PR #256, opened 2024-06-25
fix: Fix scaling for small videos (MED-628)

MarquisLP:
This PR allows the Media Player to shrink down to smaller ISF embed sizes without getting cut off. We do this by removing the minimum limit on caption font size scaling, and removing the min height for video content.

Min height is still required for audio since audio files don't have an inherent height. Min height for audio is set to 7rem to allow ISF iframes with a height of 112px to display properly without cut-off and allow enough space for the bottom UI and Play button to space themselves out without overlapping.

## Screenshots

ISF embed, using dimensions from the client case:

MED-628 - Fix Working in ISF

Vertical video:

MED-628 - Vertical Video

Wide video:

MED-628 - Wide Video

Audio in Classic Content:

MED-628 - Audio in Classic Content

Audio in Lessons:

MED-628- Audio in Lessons

Audio in NLE:

MED-628 - Audio in NLE


MarquisLP:
I've updated the Audio Player's min-height so that it uses the original height of `17rem` by default but shrinks to `90vh` if the viewpoirt is smaller than `17rem`. I've confirmed this is working in the demo page -- going to test this in ISF.

MarquisLP:
I've confirmed that with the latest changes, the Audio Player now looks as it did before, along with proper resizing in small embeds. As such, this PR is ready for re-review.

ISF embeds:
MED-628 - Audio V2 - ISF

Classic Content:
MED-628 - Audio v2 - Classic Content

Lessons:
MED-628 - Audio V2 - Lessons


NLE:
MED-628 - Audio V2 - NLE

MarquisLP:
@rcausey: I've confirmed that the cut-off from this screenshot:
MED-628 - Fix Working in ISF - UI cutoff caused by HTML Editor scrolling

was caused by scrolling in the HTML Editor. Here is the same video with the scroll position set to properly encompass the full dimensions of the Media Player:
MED-628 - Fix Working in ISF - Take 2 with no UI cutoff

---

As we discussed, I've added some logic to scale the size of the time text to prevent it from wrapping onto multiple lines. However, for very small widths (such as vertical video ISF embeds), the text will still need to wrap as this is the only way to prevent the settings gear icon from going off-screen.

MED-628 - Time font scaling

As for the cut-off of the fullscreen button on the vertical video, it was always getting cut off depending on the width of the Media Player embed. Here's the same vertical video with my local media-player repo checked out on the current main branch code (commit b0c3c056e0af1549efead5618c00e887dd822599):

MED-628 - Vertical video small size on current master code

rcausey:
> @rcausey: I've confirmed that the cut-off from this screenshot: MED-628 - Fix Working in ISF - UI cutoff caused by HTML Editor scrolling
> 
> was caused by scrolling in the HTML Editor. Here is the same video with the scroll position set to properly encompass the full dimensions of the Media Player: MED-628 - Fix Working in ISF - Take 2 with no UI cutoff
> 
> As we discussed, I've added some logic to scale the size of the time text to prevent it from wrapping onto multiple lines. However, for very small widths (such as vertical video ISF embeds), the text will still need to wrap as this is the only way to prevent the settings gear icon from going off-screen.
> 
> MED-628 - Time font scaling
> As for the cut-off of the fullscreen button on the vertical video, it was always getting cut off depending on the width of the Media Player embed. Here's the same vertical video with my local media-player repo checked out on the current main branch code (commit [b0c3c05](https://github.com/BrightspaceUILabs/media-player/commit/b0c3c056e0af1549efead5618c00e887dd822599)):
> 
> MED-628 - Vertical video small size on current master code

Thanks Mark!

rcausey (approved):

PR #269, opened 2024-10-16
MED: 1119: [JS Error] _getDownloadLink srcUrl is sometimes undefined

Syeda-Azim:


rcausey:
@Syeda-Azim Just curious: are you able to reproduce the error? Do we know why it's sometimes undefined?

Syeda-Azim:
> @Syeda-Azim Just curious: are you able to reproduce the error? Do we know why it's sometimes undefined?

I could reproduce the error in the media-player's demo by enabling [`allowDownload`](https://github.com/BrightspaceUILabs/media-player/blob/main/media-player.js#L1179C13-L1179C26) attribute. When [multiple qualities are enabled using <source>](https://github.com/BrightspaceUILabs/media-player?tab=readme-ov-file#multiple-qualities-using-source) tags, we do not set `src` on `d2l-labs-media-player`, which leads to the undefined error.

rcausey (approved):

PR #270, opened 2024-10-22
MED-1119: [JS Error] srcUrl is sometimes undefined 

Syeda-Azim:

PR #274, opened 2024-12-04
Fixed regression for console errors if legacy videos / audio are loaded

MichaelKovalchuk:
This fixes the issue reported here: https://github.com/BrightspaceUILabs/media-player/pull/270#issuecomment-2517989292

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -1184,6 +1184,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 
 	_getDownloadLink() {
 		const srcUrl = this._getCurrentSource();
+		if (!srcUrl) return '';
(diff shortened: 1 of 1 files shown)


media-player.js
```diff
@@ -1184,6 +1184,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 
 	_getDownloadLink() {
 		const srcUrl = this._getCurrentSource();
+		if (!srcUrl) return '';
```
MichaelKovalchuk, line 1187: safer to return empty string opposed to `null`

dlockhart (approved):
Thanks!

jorge-ramirez-arredondo (approved):

PR #283, opened 2025-02-13
BSPACE-163-add-image-support

SLeAsian:
[BSPACE-163](https://desire2learn.atlassian.net/browse/BSPACE-163)
- Image support is now enabled in aws bedrock. It currently only supports base64 encoded png photos.


TTanisha (approved):
👍

PR #284, opened 2025-02-13
BSPACE - 164 collect data on button press

TTanisha:
# BSPACE - 164
Connect all data needed to send to AWS using UI "send" button

## PR
When the "Send" button is pressed, a timestamp, screenshot, current transcript, and user query are collected and prepared to be sent off.

## Jira
https://desire2learn.atlassian.net/browse/BSPACE-164

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -172,7 +172,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 				border: 1px solid black;
 				display: flex;
 				flex-direction: column; /* Stack items from top to bottom */
-				height: 100%;
+				height: 100vh;
 				overflow-y: auto; /* Enables vertical scrolling when needed */
 				padding: 10px;
 			}
@@ -698,6 +698,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 		this._hovering = false;
 		this._hoveringMediaControls = false;
 		this._loading = false;
+		this._mediaCanvas = { init:false },
 		this._message = {
 			text: null,
 			type: null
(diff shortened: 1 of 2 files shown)


Shazil-R (approved):

media-player.js
```diff
@@ -1059,27 +1069,20 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
...
-
-		this._snapshot.time = this._getTimestamp();
-		this._snapshot.ctx.drawImage(this._snapshot.video, 0, 0, this._snapshot.canvas.width, this._snapshot.canvas.height);
-		this._snapshot.image = this._snapshot.canvas.toDataURL('image/jpeg');
-
-		return this._snapshot;
+	getScreenshot() {
+		if (!this._mediaCanvas.init) { this.initializeCanvas(); }
+		this._mediaCanvas.ctx.drawImage(this._media, 0, 0, this._mediaCanvas.canvas.width, this._mediaCanvas.canvas.height);
+		return this._mediaCanvas.canvas.toDataURL('image/jpeg');
```
SLeAsian, line 1075: Noticed this is jpeg. I will change aws photo format to jpeg instead then.

media-player.js
```diff
@@ -1047,6 +1048,15 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
...
 	}
 
+	createSnapshot(userQuery) {
+		return {
+			timestamp: this._getTimestamp(),
+			query: userQuery,
+			image: this.getScreenshot(),
+			transcript: this._getTruncatedTranscript()
+		};
+	}
```
SLeAsian, line 1058: Does this need to perhaps need async?

TTanisha: getScreenshot doesn't seem to need async as far as I know, but we can always change that later if it turns out that we do.

PR #291, opened 2025-02-19
BSPACE-166: Improve Chatbox UI

TTanisha:
# Improve Chatbox UI

## Jira
https://desire2learn.atlassian.net/browse/BSPACE-166

## Description
Made changes to the organization and layout of the media and chatbot elements. Also fixed an issue where the media bar would drop below the video.

## Future
The video and chatbox will keep the same scale in the X-dimension when adding/removing the chatbox but the Y-scale will shrink with the video.
I'm not entirely sure which part is causing the entire section to shrink but this can be looked at later.


media-player.js
```diff
@@ -1320,7 +1332,7 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 
 		try {
 			const response = await generateResponse(snapshot);
-			this._chatLog += DOMPurify.sanitize(`<p><b>Bot:</b> ${response.body}</p>`);
+			this._chatLog += DOMPurify.sanitize(marked(`<b>Bot:</b> ${response.body}`));
```
Shazil-R, line 1335: marked() automatically adds a paragraph tag, so we don't need to put one around the whole thing anymore

TTanisha:
Things to fix:

- Some monitors may have incorrect UI placement
- size/aspect ratio changes (should match visual diff)

PR #295, opened 2025-03-06
Fix: Aspect Ratio Overrides Max Height Constraint. 

NicholasFlorian:
This PR is needed to fix https://desire2learn.atlassian.net/browse/VOY-3077.

The line of code on line 808, manually setting the aspect-ratio has been overriding the max height even with `--d2l-labs-media-player-video-max-height` properly defined. This has led to overflow (especially with vertical content) 

I tested out a few components without this line of code and found that rendering worked for at least all our team's use cases. Although I am not sure if this line is needed for some specific use cases: vdiffs for this repo passed correctly without this line of code as well. 

before_720
after_720

### DIFF
--- a/media-player.js
+++ b/media-player.js
@@ -105,7 +105,6 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 			_hovering: { type: Boolean, attribute: false },
 			_loading: { type: Boolean, attribute: false },
 			_maintainHeight: { type: Number, attribute: false },
-			_mediaContainerAspectRatio: { type: Object, attribute: false },
 			_message: { type: Object, attribute: false },
 			_muted: { type: Boolean, attribute: false },
 			_playing: { type: Boolean, attribute: false },
@@ -165,7 +164,6 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 				max-height: var(--d2l-labs-media-player-video-max-height, 100vh);
 				min-height: 100%;
 				position: relative;
-				width: 100%;
 			}
 
 			#d2l-labs-media-player-video-poster {
@@ -671,8 +669,6 @@ class MediaPlayer extends InternalDynamicLocalizeMixin(RtlMixin(LitElement)) {
 		this._volume = 1;
 		this._webVTTParser = new window.WebVTTParser();
 		this._playRequested = false;
-		this._mediaContainerAspectRatio = {
-		};
 		this.afterCaptions = [];
 		this.beforeCaptions = [];
 	}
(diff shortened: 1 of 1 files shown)


eKoopmans:
Hey Nick, could you add that vertical demo to this repo? Could help with future regressions.

media-player.js
MarquisLP, line 1: I believe the lines that were removed in this file were related to an issue we were having in ISF -- see the "ISF Cutoff Fix" section in #185. Can you please verify that Audio and Video in ISF is still working as intended? (i.e. The player controls should still be visible and usable, albeit with a slight cut off on the bottom.)

eKoopmans:
Removing the aspect ratio is also maybe not ideal - for instance media can end up stretched horizontally with black bars:

image

I'm not sure what we want. Maybe a way to set a minimum aspect ratio like `4/3` (not a max height), to avoid videos becoming too tall?

media-player.js
eKoopmans: Do you have a way to reproduce the ISF Cutoff Fix issue, or a test case for it?

MarquisLP:
> Removing the aspect ratio is also maybe not ideal - for instance media can end up stretched horizontally with black bars:
> I'm not sure what we want. Maybe a way to set a minimum aspect ratio like 4/3 (not a max height), to avoid videos becoming too tall?

Good catch -- we've had client cases in the past involving unintended stretching and black bars. If the changes in this PR cause the Media Player to no longer respect the source video's aspect ratio, that would be a regression. Because of this, I think we'll need to explore some other ways to fix the issue from VOY-3077.

MarquisLP:
A fix for this issue should respect both the aspect ratio of the video as well as the height boundaries of the viewport. We understand that such a fix might be challenging to implement as well as verify for all use-cases of the Media Player, so if you are able to come up with a solution, the Media Team can lend some support in verifying it. Alternatively, a Media Player defect could be filed and our team will prioritize and address it according to our current timelines.

eKoopmans:
I chatted with Nick and the black bars I called out only appear if the component hits the max-height limit - so e.g. if the component is `width: 100%`, it'll do the correct aspect ratio for as long as it can, but once it hits the max height it'll start adding those black bars.

That's probably the "correct" behaviour, if we want to continue having the concept of a max height.

MarquisLP:
> I chatted with Nick and the black bars I called out only appear if the component hits the max-height limit - so e.g. if the component is width: 100%, it'll do the correct aspect ratio for as long as it can, but once it hits the max height it'll start adding those black bars.

> That's probably the "correct" behaviour, if we want to continue having the concept of a max height.

I see, yes this sounds like the intended behaviour for respecting both max height and the aspect ratio. Looking back through past PRs, I've found one screenshot showing that this was the agreed-upon behaviour for vertical videos with a max height smaller than the allotted width:

image

So if this is the case, and it's been confirmed that the ISF cut-off issue noted above has not been re-introduced, we can most likely proceed and then the Media Team will provide some help in verifying the Media Player in other parts of the LMS.

MarquisLP (approved):
Synced up with Nick and confirmed via screenshare that the ISF issue was not re-introduced. So this looks good to me. More thorough testing will be performed during Media Team QA.

PR #297, opened 2025-03-19
VOY-3077: Revert "Fix: Aspect Ratio Overrides Max Height Constraint. "

MichaelKovalchuk:
Reverts BrightspaceUILabs/media-player#295

The issue was found that breaks the Media Player transcript viewer rendering in multiple places.

We can not risk deploying this change to production without fixing the transcript viewer. 
For now, we've decided to revert the changes caused by #295  

Possible fix could be, keeping the aspect-ration calculation as-is with height and width adjustments based on the `--d2l-labs-media-player-video-max-height` property.


MarquisLP:
To provide some additional info on the Transcript Viewer issue: when the "View Transcript" button is clicked, the Media Player disappears completely.

image

MarquisLP (approved):

eKoopmans (approved):

